PR #3, opened 2025-07-09
Create langchain-oracledb Package for Oracle AI Vector Search

fileames:
This PR moves the existing Oracle AI Vector Search integrations with LangChain to a new package named _**langchain-oracledb**_ under _libs/oracledb_

The classes include:
- **OracleDocLoader**, **OracleAutonomousDatabaseLoader**: Load the documents from various sources
- **OracleSummary**: Summarize content within/outside the database
- **OracleEmbeddings** : Generate embeddings within/outside the database 
- **OracleTextSplitter**: Chunk according to different requirements using Advanced Oracle Capabilities
- **OracleVS**: Store, index, and query vectors 

More information on these classes can be found on [Oracle AI Vector Search End-to-End Demo Guide](https://github.com/langchain-ai/langchain/tree/master/cookbook/oracleai_demo.ipynb).

Additionally, for OracleVS, following features are added:
- Database filtering
- Async functions


libs/oracledb/langchain_oracledb/document_loaders/oracleadb_loader.py
```diff
@@ -0,0 +1,134 @@
+# Copyright (c) 2025 Oracle and/or its affiliates.
```
xsai9101, line 1: What is the reason of integrating these? If I understand right, all existing users of these packages will be impacted of importing. And you don't make any modification to existing public document, how do they know what should be the right import after this?

fileames: We will make the changes to the public LangChain documents after these packages are released. The users will get depreciation warnings. 
The [deprecated decorator](https://github.com/langchain-ai/langchain/blob/a04131489eb22b239f47bd85ba7b099022259659/libs/core/langchain_core/_api/deprecation.py#L71) from LangChain to mark as deprecated. Examples [here.](https://github.com/search?q=repo%3Alangchain-ai%2Flangchain-community+deprecated%28&type=code)

libs/oracledb/langchain_oracledb/embeddings/oracleai.py
```diff
@@ -0,0 +1,191 @@
...
+
+import oracledb
+
+logger = logging.getLogger(__name__)
+
+"""OracleEmbeddings class"""
+
+
+class OracleEmbeddings(BaseModel, Embeddings):
+    """Get Embeddings"""
```
cjbj, line 33: Use a single # for all single line comments throughout the code base.

cjbj (changes requested):

libs/oracledb/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+SQL_QUERY = "select prod_id, time_id from sh.costs fetch first 5 rows only"
+
+doc_loader_1 = OracleAutonomousDatabaseLoader(
+    query=SQL_QUERY,
+    user=s.USERNAME,
+    password=s.PASSWORD,
+    schema=s.SCHEMA,
+    config_dir=s.CONFIG_DIR,
+    wallet_location=s.WALLET_LOCATION,
```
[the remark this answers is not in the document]
fileames: Hi @xsai9101 , when you have a moment, could you please take a look at this and the other comments related to `OracleAutonomousDatabaseLoader`?

libs/oracledb/tests/unit_tests/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2023 Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
cjbj: If they are modified in this PR, then they need to have either "2025" or "2023, 2025" depending on when they were created.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+            logger.exception("An unexpected error occurred: {}".format(e))
+            raise RuntimeError("Unexpected error: {}".format(e)) from e
+
+    return cast(T, wrapper)
+
+
+def _table_exists(connection: Connection, table_name: str) -> bool:
+    try:
+        with connection.cursor() as cursor:
+            cursor.execute(f"select 1 from {table_name} where ROWNUM < 1")
```
cjbj, line 168: Why mix lower & upper case?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+    return len(version_parts) < len(target_parts)
+
+
+@_handle_exceptions
+def _index_exists(connection: Connection, index_name: str) -> bool:
+    # check if the index exists
+    query = """
+        SELECT index_name 
+        FROM all_indexes 
+        WHERE upper(index_name) = upper(:idx_name)
```
cjbj, line 211: Does this cope with mixed case names?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+    else:
+        logger.info("Index already exists...")
+
+
+@_handle_exceptions
+def drop_table_purge(client: Any, table_name: str) -> None:
+    """Drop a table and purge it from the database.
+
+    Args:
+        client: The OracleDB connection object.
```
cjbj, line 604: Use 'oracledb' connection object I.e. with the correct, lowercase python-oracledb module name.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+    Raises:
+        RuntimeError: If an error occurs while dropping the table.
+    """
+
+    async def context(connection: Any) -> None:
+        if await _atable_exists(connection, table_name):
+            with connection.cursor() as cursor:
+                ddl = f"DROP TABLE {table_name} PURGE"
+                await cursor.execute(ddl)
+            logger.info("Table dropped successfully...")
```
cjbj, line 640: Logging would be better if the table name was included.  Or is this a security risk?

libs/oracledb/langchain_oracledb/embeddings/oracleai.py
```diff
@@ -0,0 +1,191 @@
...
+                chunks.append(json.dumps(chunk))
+
+            vector_array_type = self.conn.gettype("SYS.VECTOR_ARRAY_T")
+            inputs = vector_array_type.newobject(chunks)
+            cursor.execute(
+                "select t.* "
+                + "from dbms_vector_chain.utl_to_embeddings(:content, "
+                + "json(:params)) t",
+                content=inputs,
+                params=json.dumps(self.params),
```
cjbj, line 129: Can this bind using DB_TYPE_JSON to avoid conversion overhead, and make use of Oracle's JSON format?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
+# Copyright (c) 2025 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
```
cjbj, line 3: Code comments describing each file's purpose would be good.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+            ]
+
+        connection = _get_connection(self.client)
+        if connection is None:
+            raise ValueError("Failed to acquire a connection.")
+        with connection.cursor() as cursor:
+            if self.json_insert_mode == "json":
+                cursor.setinputsizes(None, None, self.json_type, None)
+            cursor.executemany(
+                f"INSERT INTO {self.table_name} (id, embedding, metadata, "
```
cjbj, line 1120: Maybe because I haven't stepped through the code in a debugger, but it seems strange to me the code isn't working with the VECTOR datatype for embeddings.

cjbj (changes requested):

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+                await cursor.execute(drop_query)
+                logger.info(f"Index {index_name} has been dropped.")
+        else:
+            logger.exception(f"Index {index_name} does not exist.")
+
+    await _handle_context(client, context)
+    return
+
+
+def get_processed_ids(
```
sudarshan12s, line 697: Are these ids supposed to be unique as they are placed in primary column.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+    ) -> OracleVS:
+        (
+            client,
+            table_name,
+            distance_strategy,
+            query,
+            params,
+        ) = OracleVS._from_texts_helper(texts, embedding, metadatas, **kwargs)
+
+        drop_table_purge(client, table_name)
```
sudarshan12s, line 1886: Can you please clarify my understanding ... I have seen in javascript that user calls from_texts equavelent in batches. Like for a pdf, if its divided into 1k chunks, this function is called with 250 chunks in each call. But this drop_table_purge was removing the older 250 chunks, i had to comment for my testing. Should this be done based on some options ,.. ? other vector stores do same way?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+    unique_id = str(uuid.uuid4()).replace("-", "")
+    return f"{base_name}_{unique_id}"
+
+
+def _get_table_dict(embedding_dim: int) -> Dict:
+    cols_dict = {
+        "id": "RAW(16) DEFAULT SYS_GUID() PRIMARY KEY",
+        "text": "CLOB",
+        "metadata": "JSON",
+        "embedding": f"vector({embedding_dim}, FLOAT32)",
```
sudarshan12s, line 268: do we plan to add support for other storage types like float64/ int8,.. or flexible dimensions (*). May be we enhance the variable options later and consider them from user..

libs/oracledb/tests/unit_tests/__init__.py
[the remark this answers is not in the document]
fileames: They were not modified, just copied from oci package to follow the same structure. Is it okay to leave 2023 if not modified?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
fileames: I agree that this check was not working very well. I changed to better handle mixed case. 
Also, his check would not let users create an index if there was an index with a same name in another table and do not give any errors or logs. I added table_name to the check, now users will get `oracledb.exceptions.DatabaseError: ORA-00955: name is already used by an existing object` when they try to create indexes for different tables with the same name. New tests added for it `test_index_table_case` and `test_index_table_case_async`

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
fileames: I did not understand the comment completely, but I removed the insert_mode since we only want to support new versions. Do you think the new version check is what we should have?

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+@_handle_exceptions
+def _create_table(connection: Connection, table_name: str, embedding_dim: int) -> None:
+    cols_dict = _get_table_dict(embedding_dim)
+
+    if not _table_exists(connection, table_name):
+        with connection.cursor() as cursor:
+            ddl_body = ", ".join(
+                f"{col_name} {col_type}" for col_name, col_type in cols_dict.items()
+            )
+            ddl = f"CREATE TABLE {table_name} ({ddl_body})"
```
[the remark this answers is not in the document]
fileames: I believe we use the `_table_exists` function for logging.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
fileames: Yes, we expect them to be unique.

libs/oracledb/langchain_oracledb/utilities/oracleai.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns:
+            List of summary text, one for each input doc.
+        """
+
+        if docs is None:
+            return []
+
+        results: List[str] = []
+        try:
+            oracledb.defaults.fetch_lobs = False
```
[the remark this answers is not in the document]
fileames: @forkhroy

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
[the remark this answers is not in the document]
sudarshan12s: I mean using this syntax to avoid extra round trip check  : `CREATE TABLE IF NOT EXISTS ${tableName}`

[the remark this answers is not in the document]
sudarshan12s: since this  tablename  is taken from user , we need to quote to avoid sql injections.

something like this:
`f'"{tableName.strip().replace(\'"\', \'""\')}"'
`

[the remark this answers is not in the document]
sudarshan12s: same for all user inputs like index name which are used to form sql string

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
sudarshan12s: we use json.dumps(metadata) to convert to string. we can directly pass JSON value by using DB_TYPE_JSON  .
`
https://python-oracledb.readthedocs.io/en/latest/user_guide/json_data_type.html
`


If we convert to string in JSON column, then we loose benefits of JSON text search etc (when we do hybrid search later... )

sudarshan12s: Since you passed `array.array("f", embedding),` DB_TYPE_VECTOR is not needed. client side vector image is constructed.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -0,0 +1,1929 @@
...
+        embedding: List[float] = []
+        if isinstance(self.embedding_function, Embeddings):
+            embedding = await self.embedding_function.aembed_query(query)
+        docs_and_scores = await self.asimilarity_search_by_vector_with_relevance_scores(
+            embedding=embedding, k=k, filter=filter, **kwargs
+        )
+        return docs_and_scores
+
+    @_handle_exceptions
+    def _get_clob_value(self, result: Any) -> str:
```
sudarshan12s, line 1272: Is this function needed? 
If you use output type handler as i suggested in one of the comment and pass in execute , we get string value from clob directly.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
fileames: `json.dumps(metadata)` was coming from the earlier implementation where we wanted to support older versions. It is now removed from the code as we require new versions in this library.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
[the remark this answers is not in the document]
sudarshan12s: Afaik, it is not there. If it is important to log this, then It might be needed. It causes extra trip though with current code..

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
sudarshan12s: Thanks

libs/oracledb/langchain_oracledb/embeddings/oracleai.py
forkhroy: utl_to_embedding()'s params type is json, so i think, yes you can use db_type_json to avoid conversion overhead. however, params is not a big json doc, i don't think it has any significant impact though.

libs/oracledb/langchain_oracledb/utilities/oracleai.py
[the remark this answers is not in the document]
cjbj: @forkhroy  `oracledb.defaults.fetch_lobs` is global, and will affect any other user of python-oracledb in the app, which is not desirable.  The type handler can be confined to a connection or cursor, so it won't affect other users.

YouNeedCryDear (approved):
Probably need to add the test, build and release workflow in the following PR.

PR #8, opened 2025-07-30
langchain_oci bugfix: include ToolMessages in chat history

streamnsight:
This PR fixes a bug in the langchain integration as ToolMessages are NOT included in the chat history. 
This leads to issues with follow-up questions where the ToolMessage content is not available in the history context for the LLM to infer from.

HOWEVER, note that there is another issue with the GenAI inference service that currently would make this bugfix FAIL with Cohere Command R/R+ (`cohere.command-r-plus-08-2024`) and Command A (`cohere.command-a-03-2025`) models, in the case where a HumanMessage follows a ToolMessage.
This is a specific use-case when using the `@tool(return_direct=True)` annotation for the tool to return direct (i.e. without another pass to the LLM to rephrase the content of the tool output). This error is not currently raised because the ToolMessages are being stripped out, so it is never the case.

With this bugfix, and in the conditions above, we get:

```
oci.exceptions.ServiceError: {'target_service': 'generative_ai_inference', 'status': 400, 'code': '400', 'opc-request-id': '', 'message': '{"message":"invalid request: cannot specify message if the last entry in chat history contains tool results"}
```

This secondary issue is being worked on by OCI and should be resolved by mid August, so this fix can be safely applied after that date for all use-cases.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -51,6 +51,7 @@ from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResu
 from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_function
+from oci.generative_ai_inference import models
 from pydantic import BaseModel, ConfigDict
 
 from langchain_oci.llms.oci_generative_ai import OCIGenAIBase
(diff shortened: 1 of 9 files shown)


libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -51,6 +51,7 @@
 from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_function
+from oci.generative_ai_inference import models
```
YouNeedCryDear, line 54: I think we already have the models imported inside provider. It is a practice recommended by LangChain.

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -362,6 +363,19 @@ def messages_to_oci_params(
                         message=msg_content, tool_calls=tool_calls
                     )
                 )
+            elif isinstance(msg, ToolMessage):
+                oci_chat_history.append(
+                    self.oci_chat_message[self.get_role(msg)](
+                        tool_results=[
+                            models.CohereToolResult(
```
YouNeedCryDear, line 370: Those two lines could be replaced by self.oci_tool_result and self.oci_tool_call

YouNeedCryDear (changes requested):

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
streamnsight: fixed

YouNeedCryDear (approved):

Issue #10, opened 2025-08-01
ToolMessages in chat history - discrepancies between Command R/R+ and Command A


streamnsight:
For context, let's state the current situation:
1) There are 2 Cohere APIs: v1 and v2. At this time (08-2025), the OCI SDK, leveraged by this integration, only uses V1. The `langchain_cohere` integration uses v2.
2) The integration strips `ToolMessages` from chat history. This, to me, is a bug: `ToolMessage` content usually contains useful context that may be needed for follow-up questions.
3) Command R and R+ models do not require a `tool_result` following a `tool_call`. It does not error on that, so when ToolMessages are stripped away (as it is now), the API doesn't complain.
4) Command A, however, now **requires** a `tool_result` matching a `tool_call` when a `tool_call` is present in history, so tool calls fails with Command A because `ToolMessages` are stripped from history. This fails with the error: 
    ```
    "message":"invalid request: An assistant message with \'tool_calls\' must be followed by tool messages responding 
    to each \'tool_call_id\'.     The following tool_call_ids did not have response messages: "
    ```
    **-> To make this integration work with Command A tool calls, we need to pass the `ToolMessages` in history.**

5) When adding the `ToolMessages` in history, for Command R, R+ and A, the integration works when the history is like:
    ```
    [
      HumanMessage("give me the list of users that match X"),
      AIMessage("I will use the tool 'list_users' with parameters 'criteria' = X"),
      ToolMessage("I found the following users: \n- John Does (id 123) \n- Jane Doe (id 345)")
    ]
    ```
    where the AI is now interpreting the ToolMessage and providing a response (i.e. no `return_direct` flag)

Now,

6) LangChain has a feature for tools which is to return directly (i.e. without an extra LLM call to interpret the `tool_result`). This is activated by providing the `@tool` decorator with the parameter `@tool(return_direct=True)`.
In that case, and in the case of a chat conversation, we then typically have a history that looks like:
    ```
    [
      HumanMessage("give me the list of users that match X"),
      AIMessage("I will use the tool 'list_users' with parameters 'criteria' = X"),
      ToolMessage("I found the following users: \n- John Does (id 123) \n- Jane Doe (id 345)"),   # <--- ToolMessage returned directly
      HumanMessage("give me the contact info for John Doe"),  # <-- follow up HumanMessage
    ]
    ```
    In this case, if the ToolMessage had been removed, the follow up question from the human could not be answered. This is why I say it's a bug, and this is why Command A requires those.

[...]

PR #11, opened 2025-08-01
create dummy AIMessage repeating ToolMessage

streamnsight:
create dummy AIMessage repeating ToolMessage results to circumvent the issue with ToolMessages being stripped out.

See #10

### DIFF
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -365,9 +365,12 @@ class CohereProvider(Provider):
 
         # Process current turn messages in reverse order until a HumanMessage
         current_turn = []
-        for message in reversed(messages):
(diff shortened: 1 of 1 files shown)


streamnsight:
@YouNeedCryDear 
can we merge this now?

YouNeedCryDear (approved):

streamnsight:
@YouNeedCryDear 
What's the plan for propagating this to the langchain_community repo? are we keeping both, are we deprecating the community version, should I open a similar PR there in the meantime?
Thanks for your input

YouNeedCryDear:
> @YouNeedCryDear What's the plan for propagating this to the langchain_community repo? are we keeping both, are we deprecating the community version, should I open a similar PR there in the meantime? Thanks for your input

I think we should deprecate the langchain community version. Even though it is there, no one seems to be actively contributing to that repo any more. @streamnsight

PR #14, opened 2025-08-12
Use MetaProvider for xai and openai models.

qiuosier:
As part of ongoing enhancements, the OCI Generative AI Service is introducing additional models from xAI and OpenAI. These new models adhere to the same API specifications as the Meta models; however, the existing implementation does not automatically derive the associated `provider` for xAI or OpenAI models.

The `MetaProvider` provider can be used for xAI and OpenAI models provided by the OCI Generative AI service.
This PR adds a simple change to automatically use `MetaProvider` when user is using xAI or OpenAI models.

### DIFF
--- a/libs/oci/langchain_oci/llms/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/llms/oci_generative_ai.py
@@ -269,6 +269,8 @@ class OCIGenAI(LLM, OCIGenAIBase):
         return {
             "cohere": CohereProvider(),
             "meta": MetaProvider(),
+            "xai": MetaProvider(),
+            "openai": MetaProvider(),
         }
 
     @property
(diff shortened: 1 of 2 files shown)


libs/oci/langchain_oci/llms/oci_generative_ai.py
```diff
@@ -202,14 +208,16 @@ def _identifying_params(self) -> Mapping[str, Any]:
     def _get_provider(self, provider_map: Mapping[str, Any]) -> Any:
         if self.provider is not None:
             provider = self.provider
+        elif self.model_id.startswith(CUSTOM_ENDPOINT_PREFIX):
```
YouNeedCryDear, line 211: The order might need to change. You have to check the self.model_id is None first, otherwise, this line will raise exception.

YouNeedCryDear (changes requested):

PR #19, opened 2025-08-15
Remove unneeded deps

streamnsight:
There are several dependencies that should be optional (ads, langchain-openai)
and there are several version requirements that are very stringent and require old apps to be upgraded for lots of packages, making it complicated to switch from the langchain-community version seamlessly

### DIFF
--- a/libs/oci/pyproject.toml
+++ b/libs/oci/pyproject.toml
@@ -12,8 +12,8 @@ license = "UPL"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0"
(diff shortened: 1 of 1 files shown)


streamnsight:
@YouNeedCryDear 
removing unneeded deps and losening current required versions as they do not need to be so stringent, nad make it hard to migrate smoothly.

streamnsight:
@YouNeedCryDear 
Can we get this moving please? 
I have been needing to pack the wheel in my projects, and would love to be able to pull from pypi instead.
Let me know if this needs any changes.
Thanks

YouNeedCryDear (approved):

PR #21, opened 2025-08-18
Filter Bindings + OracleEmbeddings Optimization

fileames:
This PR introduces two key improvements:

- **Filter bindings:** Values used in filter operations are now passed as bind variables, improving security and reducing the risk of SQL injection.
- **`OracleEmbeddings` Optimization:** When using `OracleEmbeddings` with `OracleVS`, embeddings are no longer extracted and returned to the Python client. Instead, they are computed and used directly within the database, improving efficiency and reducing data transfer overhead.

### DIFF
--- a/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
+++ b/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
@@ -87,7 +87,7 @@ COMPARISON_MAP = {
     "$not": "",
 }
 
-# pperations that may need negation
+# operations that may need negation
 NOT_OPERS = ["$nin", "$not", "$exists"]
 
 


libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -78,30 +80,38 @@ def _convert_oper_to_sql(oper: str) -> str:
...
-    if isinstance(value, str):
-        value = f'"{value}"'
-    return f"JSON_EXISTS(metadata, '$.{key}?(@ {oper} {value})')"
+
+    value_bind = f":value{len(bind_variables)}"
+    bind_variables.append(value)
+
+    return (
+        f"JSON_EXISTS(metadata, "
+        f"'$.{key}[*]?(@ {oper} $val)' "
```
sudarshan12s, line 93: Can you confirm on [*]
[*] means it checks contains right? But we are checking for key = value or key > value but not key contains value .. 


like metadata has `"tags": ["db", "oracle", "json"]`, 
if user gives key: tags,  oper:EQ , value : oracle , It would work which should have failed because it contains other values db and json..

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -78,30 +80,38 @@ def _convert_oper_to_sql(oper: str) -> str:
...
+        f'PASSING {value_bind} AS "val")'
+    )
 
 
-def _generate_where_clause(db_filter: Union[FilterCondition, FilterGroup]) -> str:
+def _generate_where_clause(
+    db_filter: FilterCondition | FilterGroup, bind_variables: List[str]
```
sudarshan12s, line 99: we might still need to support simple filter like blow for and (SODA filter allows this apart from $and keyword. This can be considered for future i think... 

`{ category: "books", price: 10 }`

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
fileames: Again, talking for the new format, equality is meant to be used with scalars. If size known users can use indexing {"tags[0]" : "db", "tags[1]" : "oracle", "tags[2]" : "json"}. Maybe we can add a $size operator to make sure they can implement equality with lists.

fileames: I added list and dict equality using JSON_QUERY and JSON_EQUAL in the last commit.

libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
```diff
@@ -78,30 +80,38 @@ def _convert_oper_to_sql(oper: str) -> str:
     return oper_map.get(oper, "==")
 
 
-def _generate_condition(condition: FilterCondition) -> str:
-    key = condition["key"]
+def _generate_condition(condition: FilterCondition, bind_variables: List[str]) -> str:
+    key = _quote_indentifier(condition["key"])
     oper = _convert_oper_to_sql(condition["oper"])
     value = condition["value"]
```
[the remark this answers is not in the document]
sudarshan12s: Thanks

ashutoshSce (approved):

PR #29, opened 2025-09-03
Don't swallow connection errors in the connectivity test example

cjbj:
- Removed try/except block from example python-oracledb connection code so that any connection error is shown, allowing the user to troubleshoot.
- Added a section with doc links.
- Trim trailing whitespace.


fileames (approved):
Looks good, thanks!

Issue #40, opened 2025-10-09
with_structured_output() does not work with Gemini 2.5 Models


MeerrettichMeister:
Using structured output with the new Gemini models dont work.

```py
from langchain_core.messages import HumanMessage
from langchain_oci import ChatOCIGenAI
from pydantic import BaseModel


class Formatted(BaseModel):
    answer: str
    user_request: str


oci_chat = ChatOCIGenAI(
    model_id="google.gemini-2.5-flash",
    service_endpoint="https://inference.generativeai.us-ashburn-1.oci.oraclecloud.com",
    compartment_id="compartment_id",
)


unstructured_response = oci_chat.invoke(
    [HumanMessage("Tell me something about the statue of liberty")]
)

structured_chat = oci_chat.with_structured_output(Formatted)
structured_response = structured_chat.invoke(
    [HumanMessage("Tell me something about the statue of liberty")]
)
print(unstructured_response)
print(structured_response)
```

Running the above code returns a proper response with the regular chat interface but fails with the structured output.
It either returns None or fails by the pydantic validation.

Python 3.12
langchain 0.3.27
langchain-oci 0.1.5
pydantic 2.11.7

PR #54, opened 2025-10-29
[BUG] Fix json_schema method in the with_structured_output function.

paxiaatucsdedu:
# Problem
Issue link: https://github.com/oracle/langchain-oracle/issues/40

1.` with_structured_output()` use `method="function_calling"` as default method. When calling Gemini 2.5 models, no function calling is returned so `with_structured_output() `does not work with Gemini 2.5 models.
2. If using the `json_schema` method in `with_structured_output() `, `with_structured_output() `was failing with a 400 error ("Please pass in correct format of request"). The root cause is that the code was passing a plain Python dict for response_format instead of proper OCI SDK objects that the GenericChatRequest API expects.

# Solution
1. Suggest in README that user should use `json_schema` method when the default method does not work.
2. Updated the `json_schema` method to use proper OCI SDK objects:

- ResponseJsonSchema - wraps the JSON schema with name, description, and strict mode
- JsonSchemaResponseFormat - the proper response format type expected by the API

# Testing
Now Gemini-2.5 model returns properly formatted object instead of None when using `json_schema` method in `with_structured_output() ` function.
image

### DIFF
--- a/libs/oci/README.md
+++ b/libs/oci/README.md
@@ -63,7 +63,8 @@ embeddings.embed_query("What is the meaning of life?")
 
 ### 4. Use Structured Output
 `ChatOCIGenAI` supports structured output. 
-The default method is "function_calling". If default method fails, try "json_schema" or "json_mode".
+
+<sub>**Note:** The default method is `function_calling`. If default method returns `None` (e.g. for Gemini models), try `json_schema` or `json_mode`.</sub>
 
 ```python
 from langchain_oci import ChatOCIGenAI


libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -1129,7 +1129,7 @@ def with_structured_output(
         *,
         method: Literal[
             "function_calling", "json_schema", "json_mode"
-        ] = "function_calling",
+        ] = "json_schema",
```
YouNeedCryDear: Can you double check if the langchain openai is still using function_calling as default? Changing default might introduce unexpected behavior for customer.

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -1201,19 +1201,26 @@ def with_structured_output(
                 else JsonOutputParser()
             )
         elif method == "json_schema":
-            response_format = (
-                dict(
-                    schema.model_json_schema().items()  # type: ignore[union-attr]
-                )
+            from oci.generative_ai_inference import models
```
YouNeedCryDear: do not import here, instead define an alias for the class in the init

PR #61, opened 2025-11-04
Add support for OCI OpenAI Responses API with Langchain/Langgraph

furqan-shaikh-dev:
# Summary
Adds support for OpenAI Responses API by introducing a new class `OCIChatOpenAI`. This allows users to invoke OCI Generative AI to leverage OpenAI Responses API.

# Changes
- Add new file `oci_generative_ai_responses_api.py` which implements `OCIChatOpenAI` class and OCI Auth.
- Add comprehensive unit tests covering most common langchain and langgraph scenarios.
- Add `examples` folder to provide sample usages
- Add documentation related to the change in `README.md`

# Testing
- Unit Tests: 5/5 tests passing
- Integration Tests: To be raised as a separate PR

# Breaking Changes
None - this is a new feature that is fully backward compatible.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -52,7 +52,6 @@ from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_function
 from langchain_openai import ChatOpenAI
-from oci_openai.oci_openai import _resolve_base_url
 from openai import DefaultHttpxClient
 from pydantic import BaseModel, ConfigDict, SecretStr, model_validator
 
@@ -1611,6 +1610,14 @@ class ChatOCIOpenAI(ChatOpenAI):
         base_url: Optional[str] = None,
         **kwargs: Any,
     ):
+        try:
+            from oci_openai.oci_openai import _resolve_base_url
+        except ImportError as e:
+            raise ImportError(
+                "Could not import _resolve_base_url. "
+                "Please install: pip install oci-openai"
+            ) from e
+
         super().__init__(
             model=model,
             api_key=SecretStr(API_KEY),


libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
```diff
@@ -0,0 +1,407 @@
...
+)
+from pydantic import SecretStr, model_validator
+
+API_KEY = "<NOTUSED>"
+COMPARTMENT_ID_HEADER = "opc-compartment-id"
+CONVERSATION_STORE_ID_HEADER = "opc-conversation-store-id"
+OUTPUT_VERSION = "responses/v1"
+
+
+class OCIHttpxAuth(httpx.Auth):
```
YouNeedCryDear: I think the oci_openai has been published in github. Can we use that?

YouNeedCryDear (changes requested):

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
```diff
@@ -0,0 +1,407 @@
...
+            private_key_file_location=config.get("key_file"),
+            # pass_phrase is optional and can be None
+            pass_phrase=oci.config.get_config_value_or_default(config, "pass_phrase"),
+            # private_key_content is optional and can be None
+            private_key_content=config.get("key_content"),
+        )
+        super().__init__(signer=signer)
+
+
+def get_base_url(region: str, override_url: str = "") -> str:
```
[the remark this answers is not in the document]
furqan-shaikh-dev: you mean the existing `utils` under `llms`? we have kept it at a single place since its being only used by `oci_generative_ai_responses_api` module. wdyt?

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
anbhaduri: We can only start taking dependency on oci_openai post GA when we start publishing it to pypi,

YouNeedCryDear: Why cant we leverage this? https://github.com/oracle-samples/oci-openai/tree/v0.2.1
I don't think it is a good idea to have separate Auth defined.

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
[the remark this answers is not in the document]
YouNeedCryDear: I think it is better to put into utils for extensibility, even for now it is only used for responses api.

YouNeedCryDear:
> @YouNeedCryDear wouldn't it be a lot easier to switch to the oci_openai package when making calls with external models (llama, grok, gpt...), then the whole support would be propagated from the openai original SDK, and the only support we need to implement is support for Cohere models https://github.com/oracle-samples/oci-openai

Sounds like a good idea. But this might be a huge change to the code base and I don't think the Reponses API team is able to do.

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
```diff
@@ -0,0 +1,407 @@
+import importlib
+from typing import Any, Generator, Mapping, Union
+
+import httpx
+import oci
+import requests
+from langchain_openai import ChatOpenAI
+from oci.config import DEFAULT_LOCATION, DEFAULT_PROFILE
+from openai import (
```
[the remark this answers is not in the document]
furqan-shaikh-dev: DONE

YouNeedCryDear:
> > The change looks good to me. One more question regarding the file naming and structure. Is the ChatOCIOpenAI specifically for responses API or it can also be used for chat-completion? I wonder if we should rename the file to something more generic or directly move the class under oci_generative_ai.py. @furqan-shaikh-dev
> 
> the openai API supports both. the langchain-openai tries to detects if the model selected used reasoning, and if so will automatically select the Response API, however it is also a flag you can set. in the example here: https://github.com/oracle-samples/oci-openai?tab=readme-ov-file#using-with-langchain-openai you can see how that works with langchai-openai So, if we are wrapping the functionality, we need to make sure the flag is supported.

That's why I think instead of having it under oci_generative_ai_responses_api.py, which the name is specific for responses api, we probably need to rename the file or just directly move the ChatOCIOpenAI class under oci_generative_ai.py

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
```diff
@@ -0,0 +1,165 @@
...
+
+          pip install -U langchain-oci openai langchain-openai
+
+    Attributes:
+        auth (httpx.Auth): Authentication handler for OCI
+        compartment_id (str): OCI compartment ID for resource isolation
+        conversation_store_id (str): Conversation Store Id to use when generating responses
+        region (str): OCI Region where the OCI GenAI Unified API resides
+        model (str): Name of OpenAI model to use.
+        service_endpoint (str): OCI Unified API URL to override , if it's different from base url
```
ericjy: There is difference between service_endpoint override and base_url override.

Learn more https://github.com/oracle-samples/oci-openai/pull/10#discussion_r2539631941

libs/oci/examples/chat_models/langchain_oci_openai_responses_api_example.py
```diff
@@ -0,0 +1,116 @@
...
+from oci_openai import OciSessionAuth
+from langchain_oci import ChatOCIOpenAI
+from rich import print
+
+COMPARTMENT_ID = ""
+CONVERSATION_STORE_ID = ""
+OVERRIDE_URL = ""
+REGION = ""
+MODEL = ""
+PROFILE_NAME = ""
```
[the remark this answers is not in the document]
furqan-shaikh-dev: DONE, replaced with dummy values

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
```diff
@@ -0,0 +1,165 @@
...
+from oci_openai.oci_openai import _build_base_url, _build_service_endpoint, _resolve_base_url
+from openai import DefaultHttpxClient
+
+from pydantic import SecretStr, model_validator
+
+
+API_KEY = "<NOTUSED>"
+COMPARTMENT_ID_HEADER = "opc-compartment-id"
+CONVERSATION_STORE_ID_HEADER = "opc-conversation-store-id"
+OUTPUT_VERSION = "responses/v1"
```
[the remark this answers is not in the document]
furqan-shaikh-dev: This is defined by langchain itself, for supporting openai responses API. Please see: https://github.com/langchain-ai/langchain/blob/525d5c0169225f4a56d732be0dad8ad5c78b7265/libs/partners/openai/langchain_openai/chat_models/base.py#L788

libs/oci/langchain_oci/chat_models/oci_generative_ai_responses_api.py
furqan-shaikh-dev: DONE

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -1445,3 +1452,157 @@ def _stream(
...
+        model: str,
+        region: str = None,
+        service_endpoint: str = None,
+        base_url: str = None,
+        **kwargs: Any,
+    ):
+        super().__init__(
+            model=model,
+            api_key=SecretStr(API_KEY),
+            http_client=get_sync_httpx_client(
```
ericjy: What's our plan for Async client? Does LangChain has that?

libs/oci/README.md
```diff
@@ -79,6 +79,32 @@ structured_llm = llm.with_structured_output(Joke)
...
+```python
+from langchain_oci import (
+    ChatOCIOpenAI,
+    OCISessionAuth,
+)
+client = ChatOCIOpenAI(
+        auth=OCISessionAuth(profile_name="MY_PROFILE_NAME"),
+        compartment_id="MY_COMPARTMENT_ID",
+        region="MY_REGION",
+        override_url="MY_OVERRIDE_URL",
```
ericjy: override_url doesn't exist. you mean service_endpoint or base_url?

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -1445,3 +1452,157 @@ def _stream(
...
+            )
+class ChatOCIOpenAI(ChatOpenAI):
+    """A custom OCI OpenAI client implementation confirming to OpenAI Responses API.
+
+    Setup:
+      Install ``openai`` and ``langchain-openai``.
+
+      .. code-block:: bash
+
+          pip install -U langchain-oci openai langchain-openai
```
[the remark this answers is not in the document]
furqan-shaikh-dev: DONE

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -1445,3 +1452,157 @@ def _stream(
...
+
+    Attributes:
+        auth (httpx.Auth): Authentication handler for OCI request signing.
+        compartment_id (str): OCI compartment ID for resource isolation
+        conversation_store_id (str): Conversation Store Id to use when generating responses
```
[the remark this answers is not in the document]
furqan-shaikh-dev: would the responses api fail if we dont provide it?

libs/oci/README.md
furqan-shaikh-dev: DONE

ericjy (approved):

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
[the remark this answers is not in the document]
furqan-shaikh-dev: Yes, it does, pushed the code, test and doc related to that. Thanks for pointing this out

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -50,12 +51,19 @@
 from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_function
-from pydantic import BaseModel, ConfigDict
+from langchain_openai import ChatOpenAI
+from oci_openai.oci_openai import _resolve_base_url
```
YouNeedCryDear: this import should go inside the class with try except block.

YouNeedCryDear (changes requested):

YouNeedCryDear (approved):

PR #70, opened 2025-11-21
[BUGS] Handle double-escaped JSON in streaming tool calls

paxiaatucsdedu:
# Problem
Streaming tool calls were not handling double-escaped JSON arguments, while the non-streaming path (convert_oci_tool_call_to_langchain) already had this fix. This caused tool call arguments to be incorrectly parsed in streaming mode.

# Solution
Applied the same double-escape handling logic to process_stream_tool_calls in both CohereProvider and GenericProvider:

```
args = tool_call["function"].get("arguments")
try:
    parsed_args = json.loads(json.loads(args))
    args = json.dumps(parsed_args)
except (json.JSONDecodeError, TypeError):
    pass
```

Logic:

- Normal JSON ('{"key": "value"}'): First parse succeeds → dict → second parse raises TypeError → keep original

- Double-escaped JSON ('"{\"key\": \"value\"}"'): First parse → string → second parse → dict → convert back to unescaped JSON

- Invalid/empty JSON: First parse raises JSONDecodeError → keep original

### DIFF
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -563,10 +563,19 @@ class CohereProvider(Provider):
             if tool_id:
                 tool_call_ids.add(tool_id)
 
+            args = tool_call["function"].get("arguments")
+            # If args is a double-escaped JSON string, parse it twice to get the original JSON  # noqa: E501
+            try:
+                parsed_args = json.loads(json.loads(args))
+                args = json.dumps(parsed_args)
+            # If args is not a double-escaped JSON string, keep it as is
+            except (json.JSONDecodeError, TypeError):
+                pass
+
             tool_call_chunks.append(
                 tool_call_chunk(
                     name=tool_call["function"].get("name"),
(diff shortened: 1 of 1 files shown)


libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -563,10 +563,19 @@ def process_stream_tool_calls(
             if tool_id:
                 tool_call_ids.add(tool_id)
 
+            args = tool_call["function"].get("arguments")
```
kirankumarjoseph: It will take more than this to fix this problem. 
When langgraph tries to consume streaming chunks and tries to create a tool call, it will fail if the parsed string is not a json and it will create an invalid tool call much before the control comes to our code.
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/ai.py#L508-L522

kirankumarjoseph: One option is to extend `AIMessageChunk` and override `init_tool_calls` to do the double parsing ourselves. Make sure you use the new class in this file instead of `AIMessageChunk`

PR #109, opened 2026-01-29
Fix tool schema constraints dropped during OCI conversion (breaks MCP + Gemini)

fede-kamel:
## Bug

`convert_to_oci_tool` strips JSON Schema constraints (enum, min/max, pattern, format) from tool definitions before sending them to the model. A tool defined with `enum: ["cpu", "memory", "disk", "network"]` gets reduced to just `type: string` — the model never sees the allowed values.

Additionally, Optional fields (Pydantic `anyOf` pattern) get converted to `type: "any"` which is not valid JSON Schema — this causes Gemini models to return 400 errors.

Root cause: the old code uses `tool.args` which goes through Pydantic's `tool_call_schema` re-generation, then only keeps `type` and `description`:

```python
# old code
p_name: {
    "type": p_def.get("type", "any"),
    "description": p_def.get("description", ""),
}
```

## Reproduction

Tool with `enum: ["cpu", "memory", "disk", "network"]`, prompt: "Check the latency metrics" (intentionally not in enum).

**main branch (broken)** — enum stripped, models hallucinate or crash:

```
meta.llama-3.3-70b-instruct:    metric_name = "latency"  (not a valid value)
google.gemini-2.5-flash:        metric_name = "latency"  (not a valid value)
google.gemini-2.5-pro:          metric_name = "latency"  (not a valid value)
google.gemini-2.5-flash-lite:   metric_name = "latency"  (not a valid value)
```

**fix branch** — enum preserved, models respect constraints:

```
meta.llama-3.3-70b-instruct:    metric_name = "network"  (closest valid value)
google.gemini-2.5-flash:        refused — "can only query cpu, memory, disk, network"
google.gemini-2.5-pro:          refused — "available metrics are: cpu, memory, disk, network"
google.gemini-2.5-flash-lite:   refused — "can only query cpu, memory, disk, network"
```

## MCP tool reproduction

Same test with a FastMCP server (`MetricName` enum + `duration_hours` with `ge=1, le=168`).

**main branch** — MCP tools broken on Gemini, hallucinated on Llama:

```
What main sends to OCI:
  metric_name:    {"type": "string", "description": ""}     ← enum stripped
  duration_hours: {"type": "any", "description": "..."}     ← anyOf becomes "any"

meta.llama-3.3-70b:      metric_name = "latency"  (hallucinated)
google.gemini-2.5-flash: 400 error  (type "any" rejected by Gemini)
google.gemini-2.5-pro:   400 error  (type "any" rejected by Gemini)
```

**fix branch** — MCP tools work correctly:

```
What fix sends to OCI:
  metric_name:    {"type": "string", "enum": ["cpu","memory","disk","network"]}
  duration_hours: {"type": "integer", "minimum": 1, "maximum": 168}

meta.llama-3.3-70b:      metric_name = "network"  (valid)
google.gemini-2.5-flash: refused — "does not support latency"
google.gemini-2.5-pro:   refused — "can check: cpu, memory, disk, network"
```

## Fix

- Use `args_schema.model_json_schema()` instead of `tool.args` to get the original schema with constraints intact
- `_sanitize_tool_property()` filters through an allowlist of standard JSON Schema keys (enum, format, min/max, pattern, items, etc.) and resolves Pydantic v2 `anyOf` patterns for `Optional[T]`
- For Cohere: `_enrich_description()` embeds constraints into the description string since CohereParameterDefinition only supports type/description/is_required

## Tests

- 22 unit tests for `_sanitize_tool_property` (enum, format, ranges, patterns, nested objects, anyOf resolution, MCP round-trips)
- 6 integration tests against live OCI (Meta Llama + Cohere): enum constraints, numeric ranges, schema verification

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/cohere.py
+++ b/libs/oci/langchain_oci/chat_models/providers/cohere.py
@@ -463,6 +463,31 @@ class CohereProvider(Provider):
         # Remove keys with None values
         return {k: v for k, v in oci_params.items() if v is not None}
 
+    @staticmethod
+    def _enrich_description(description: str, p_def: Dict[str, Any]) -> str:
+        """Embed schema constraints into the description for Cohere models.
+
+        CohereParameterDefinition only supports type, description, and
+        is_required. Rich schema metadata (enum, format, range, pattern)
+        is embedded into the description string so the LLM can still see
+        and respect these constraints.
+        """
+        parts = [description] if description else []
+        if "enum" in p_def:
+            parts.append(f"Allowed values: {p_def['enum']}")
+        if "format" in p_def:
+            parts.append(f"Format: {p_def['format']}")
+        if "minimum" in p_def or "maximum" in p_def:
+            range_parts = []
+            if "minimum" in p_def:
+                range_parts.append(f"min={p_def['minimum']}")
+            if "maximum" in p_def:
+                range_parts.append(f"max={p_def['maximum']}")
+            parts.append(f"Range: {', '.join(range_parts)}")
+        if "pattern" in p_def:
+            parts.append(f"Pattern: {p_def['pattern']}")
+        return ". ".join(parts) if parts else ""
+
     def convert_to_oci_tool(
         self,
         tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
@@ -474,6 +499,17 @@ class CohereProvider(Provider):
         or Pydantic models/callables.
         """
         if isinstance(tool, BaseTool):
+            # Use args_schema.model_json_schema() to get rich properties
+            # (enum, constraints) that tool.args loses via tool_call_schema.
+            if tool.args_schema and hasattr(tool.args_schema, "model_json_schema"):
+                schema = tool.args_schema.model_json_schema()
+                # Resolve $ref/$defs and anyOf — OCI doesn't support them
+                schema = OCIUtils.resolve_schema_refs(schema)
+                schema = OCIUtils.resolve_anyof(schema)
+                properties = schema.get("properties", {})
+            else:
+                properties = tool.args
+
             return self.oci_tool(
                 name=tool.name,
                 description=OCIUtils.remove_signature_from_tool_description(
@@ -481,14 +517,16 @@ class CohereProvider(Provider):
                 ),
                 parameter_definitions={
                     p_name: self.oci_tool_param(
-                        description=p_def.get("description", ""),
+                        description=self._enrich_description(
+                            p_def.get("description", ""), p_def
+                        ),
                         type=JSON_TO_PYTHON_TYPES.get(
                             p_def.get("type"),
                             p_def.get("type", "any"),
                         ),
                         is_required="default" not in p_def,
                     )
-                    for p_name, p_def in tool.args.items()
+                    for p_name, p_def in properties.items()
                 },
             )
         elif isinstance(tool, dict):
@@ -502,7 +540,9 @@ class CohereProvider(Provider):
                 description=tool.get("description"),
                 parameter_definitions={
                     p_name: self.oci_tool_param(
-                        description=p_def.get("description", ""),
+                        description=self._enrich_description(
+                            p_def.get("description", ""), p_def
+                        ),
                         type=JSON_TO_PYTHON_TYPES.get(
                             p_def.get("type"),
                             p_def.get("type", "any"),
@@ -524,7 +564,9 @@ class CohereProvider(Provider):
                 ),
                 parameter_definitions={
                     p_name: self.oci_tool_param(
-                        description=p_def.get("description", ""),
+                        description=self._enrich_description(
+                            p_def.get("description", ""), p_def
+                        ),
                         type=JSON_TO_PYTHON_TYPES.get(
                             p_def.get("type"),
                             p_def.get("type", "any"),
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -378,6 +378,20 @@ class GenericProvider(Provider):
         """
         # Check BaseTool first since it's callable but needs special handling
         if isinstance(tool, BaseTool):
+            # Use model_json_schema() if available to preserve json_schema_extra
+            # constraints (enum, format, etc.) that convert_to_openai_function strips
+            if tool.args_schema and hasattr(tool.args_schema, "model_json_schema"):
+                schema = tool.args_schema.model_json_schema()
+                parameters = schema
+            else:
+                as_json_schema_function = convert_to_openai_function(tool)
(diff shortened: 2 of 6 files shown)


fede-kamel:
Hi @paxiaatucsdedu, thanks for the suggestion! I ran a head-to-head comparison of the three approaches (main, this PR, and your proposal) across 14 edge-case scenarios. Here are the results:

## Scorecard

```
Scenario                    MAIN     PR  COMMENT
------------------------------------------------
01_enum_extra               FAIL   PASS     FAIL
02_range_native             FAIL   PASS     PASS
03_opt_enum_extra           FAIL   PASS     FAIL
04_opt_int_anyof            FAIL   PASS     PASS
05_pattern_native           FAIL   PASS     PASS
06_format_extra             FAIL   PASS     FAIL
07_length_native            FAIL   PASS     PASS
08_array_item_extra         FAIL   PASS     FAIL
09_nested_ref               FAIL   FAIL     PASS
10_full_mcp_combo           FAIL   PASS     FAIL
11_native_enum_ref          FAIL   FAIL     PASS
12_exclusive_native         FAIL   PASS     PASS
13_multi_optional           FAIL   PASS     PASS
14_const_extra              FAIL   PASS     FAIL
------------------------------------------------
TOTAL                       0/14   12/14    8/14
```

## What each scenario tests

| # | Scenario | What it tests |
|---|----------|---------------|
| 01 | `enum_extra` | `json_schema_extra={"enum": ["cpu", "memory", ...]}` — enum values on a simple `str` field |
| 02 | `range_native` | `Field(ge=1, le=168)` — Pydantic-native `minimum`/`maximum` on an `int` field |
| 03 | `opt_enum_extra` | `Optional[str]` + `json_schema_extra={"enum": [...]}` — anyOf resolution AND enum preservation together |
| 04 | `opt_int_anyof` | `Optional[int]` — Pydantic emits `anyOf: [{type: integer}, {type: null}]`, must resolve to `type: integer` not `type: any` |
| 05 | `pattern_native` | `Field(pattern=r"^[\w.]+@[\w.]+$")` — Pydantic-native regex pattern |
| 06 | `format_extra` | `json_schema_extra={"format": "date-time"}` — format constraint on a string field |
| 07 | `length_native` | `Field(min_length=1, max_length=100)` — Pydantic-native string length constraints |
| 08 | `array_item_extra` | `List[str]` with `json_schema_extra={"items": {"enum": [...]}}` — enum inside array items |
| 09 | `nested_ref` | Nested Pydantic `BaseModel` — `model_json_schema()` emits `$defs` + `$ref`, must be resolved |
| 10 | `full_mcp_combo` | Combined: enum (extra) + range (native) + optional enum (extra+anyOf) — the full MCP tool scenario |
| 11 | `native_enum_ref` | Python `Enum` class (`class MetricNameEnum(str, Enum)`) — produces `$defs` + `$ref` in schema |
| 12 | `exclusive_native` | `Field(gt=0, lt=1.0)` — Pydantic-native `exclusiveMinimum`/`exclusiveMaximum` |
| 13 | `multi_optional` | Three `Optional[T]` fields (`str`, `int`, `bool`) — stress-tests anyOf resolution |
| 14 | `const_extra` | `json_schema_extra={"const": "v1"}` — const value constraint |

## Root cause of the difference

The two approaches use different schema sources, and each has a blind spot:

**This PR** uses `model_json_schema()`:
- ✅ Preserves `json_schema_extra` fields (`enum`, `format`, `const`) — this is where the 6 scenarios that only this PR passes come from
- ❌ Fails on `$ref`/`$defs` schemas (native Python `Enum`, nested Pydantic models) because `_sanitize_tool_property` strips `$ref` without resolving it first (scenarios 09, 11)

**Your proposal** uses `convert_to_openai_function()`:
- ✅ Resolves `$ref`/`$defs` automatically (LangChain handles this internally) — wins on nested objects + native enums
- ❌ Strips `json_schema_extra` fields — `convert_to_openai_function()` re-generates the schema through LangChain's `tool_call_schema` which drops `json_schema_extra` metadata (scenarios 01, 03, 06, 08, 10, 14)

**Both** handle Pydantic-native constraints (`ge`/`le`, `pattern`, `min_length`/`max_length`, `gt`/`lt`) and `anyOf` resolution for `Optional[T]`.

## Next step

The ideal fix is to combine both strengths: add `resolve_schema_refs()` to this PR's pipeline (call it on the `model_json_schema()` output **before** passing to `_sanitize_tool_property()`). That would give **14/14**. I'll update the PR with this fix.

## Test script

Full reproducible test script: https://gist.github.com/fede-kamel/43f728af00792491fab5cb48b418e34c

Run with:
```bash
# from libs/oci/
poetry run pytest tests/unit_tests/chat_models/test_compare_approaches.py::test_scorecard -s
```

fede-kamel:
Updated PR with the combined approach from #109 and #118.

What changed:

GenericProvider (from #118): uses model_json_schema() instead of convert_to_openai_function() to preserve json_schema_extra constraints (enum, format, const, pattern, min/max, etc). resolve_anyof() handles Pydantic v2 Optional[T] emitting anyOf patterns that OCI models don't understand.

CohereProvider (from #109): same model_json_schema() + resolve_schema_refs() + resolve_anyof() pipeline, then _enrich_description() embeds constraints into description text since CohereParameterDefinition only supports type/description/is_required.

Shared utilities in common/utils.py:
- OCIUtils.resolve_schema_refs() — inlines $ref/$defs (nested models, Python Enums)
- OCIUtils.resolve_anyof() — resolves anyOf to concrete types (prevents type: "any" which breaks Gemini)

Both methods are used by GenericProvider and CohereProvider (DRY).

Testing:

Unit tests (27 total, all pass):
- 14 GenericProvider scenarios: enum, range, optional+enum, optional int, pattern, format, string length, array items, nested refs, combined MCP, native Python Enum, exclusive range, multi-optional, const
- 5 defensive checks: no type "any" anywhere, no $ref/$defs remaining, required fields correct, descriptions preserved, backward compat with simple tools
- 8 CohereProvider tests: optional type resolution, enum/range/pattern in descriptions, nested refs, native enum, no type "any", full MCP combo

Integration tests (9 total, all pass against real OCI API):
- test_enum_constraints_respected — Meta Llama 4 Scout, Cohere Command A, Gemini 2.5 Flash
- test_numeric_range_constraints — Meta Llama 4 Scout, Cohere Command A, Gemini 2.5 Flash
- test_schema_sent_to_oci_has_constraints — Meta Llama 4 Scout, Cohere Command A, Gemini 2.5 Flash

Gemini was the model that was failing before this fix (type: "any" in tool schemas). Now passes cleanly.

Full test suite: 142 unit tests pass, 0 failures. 156 integration tests pass (5 failures unrelated — Cohere Vision needs newer OCI SDK, Wikimedia rate limit).

Gist with the unit test file: https://gist.github.com/fede-kamel/030bea7fc7576c8a39e92c6ca95bbd14

Credit to @paxiaatucsdedu for the model_json_schema() + resolve_anyof approach in #118.

paxiaatucsdedu (approved):

PR #110, opened 2026-01-30
Add reasoning_content extraction and null guards for GenericProvider

fede-kamel:
Reasoning models like xAI Grok-3-mini and OpenAI GPT-oss return a reasoning_content field containing chain-of-thought reasoning. The OCI SDK exposes this field on the message object, but GenericProvider was silently dropping it -- the field never made it into generation_info or additional_kwargs on the AIMessage. Users of reasoning models through langchain-oci had no way to access the model's reasoning.

The same issue exists upstream in langchain-openai (https://github.com/langchain-ai/langchain/pull/34705, pending review). In both cases the code predates reasoning models and was never updated to extract the new field. The router nature of OCI GenAI makes it more visible since new model vendors with reasoning capabilities keep getting added to the platform.

The fix checks for reasoning_content on the response message and includes it in generation_info, which LangChain automatically passes through to additional_kwargs.

Also fixed two related fragility issues in GenericProvider: chat_response_to_text, chat_generation_info, and chat_tool_calls assumed choices[0] exists (crashes on empty choices), and create_usage_metadata used getattr(usage, "prompt_tokens", 0) which doesn't help when the attribute exists but is None.


libs/oci/langchain_oci/chat_models/providers/generic.py
```diff
@@ -136,12 +139,25 @@ def is_chat_stream_end(self, event_data: Dict) -> bool:
...
+        """Extract generation metadata from chat response."""
+        choices = response.data.chat_response.choices
+        finish_reason = choices[0].finish_reason if choices else None
         generation_info: Dict[str, Any] = {
-            "finish_reason": response.data.chat_response.choices[0].finish_reason,
+            "finish_reason": finish_reason,
             "time_created": str(response.data.chat_response.time_created),
         }
 
+        # Extract reasoning_content if present (OCI reasoning models).
```
paxiaatucsdedu: Maybe we can use a shorter and more concise comments here?

fede-kamel:
Addressed all feedback:

1. Simplified `chat_response_to_text` — combined null checks into a more concise flow
2. Shortened the reasoning_content comment to one line
3. Merged unit tests into `test_oci_generative_ai.py` (6 new tests)
4. Merged integration tests into `test_chat_features.py` (5 new tests)
5. Deleted standalone `test_reasoning_content*.py` files

Note: The integration tests for reasoning models now skip gracefully because the OCI SDK doesn't yet expose `reasoning_content` on the message object. The unit tests still verify the extraction logic works when the field is present.

Tests: 104 unit passed, 3 integration passed + 2 skipped (SDK limitation)

paxiaatucsdedu (approved):

PR #111, opened 2026-01-30
Fix Gemini parallel tool calling — flatten parallel calls into sequential pairs

fede-kamel:
Gemini models through OCI GenAI fail with a 400 error when the model makes parallel tool calls (multiple function calls in one turn). The error is:

  "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn."

This is a known Gemini API constraint — each turn must have exactly one function call matched with one function response. When LangChain sends one AIMessage with N tool_calls followed by N ToolMessages, the OCI/Gemini backend rejects it.

This is actively being reported across multiple projects right now:
- google-gemini/gemini-cli#16906 (Jan 17, 2026)
- google-gemini/gemini-cli#16415 P1 (Jan 12, 2026)
- google-gemini/gemini-cli#16309 (Jan 10, 2026)
- BerriAI/litellm#17949 (Dec 14, 2025)

The fix adds _flatten_parallel_tool_calls() to GenericProvider. It rewrites the message history so that one AIMessage with N tool_calls + N ToolMessages becomes N sequential (AIMessage, ToolMessage) pairs. Each pair has exactly 1 call and 1 response, which Gemini accepts. It only activates for google.* model IDs — all other models are unaffected.

Reproduced on main without the fix:
- Gemini returns 2 parallel tool calls (get_weather + get_time)
- Sending tool results back fails with 400 INVALID_ARGUMENT

With the fix:
- Same scenario, same parallel calls
- Tool results sent back successfully
- Response: "The weather in New York City is sunny and 72F. The time is 3:00 PM."

Changes:
- langchain_oci/chat_models/providers/generic.py: add _flatten_parallel_tool_calls() and model_id routing in messages_to_oci_params()
- langchain_oci/chat_models/oci_generative_ai.py: pass model_id to messages_to_oci_params()

Tests:
- 12 unit tests (flattening logic, model routing, passthrough)
- 11 integration tests (2/3/4 parallel calls, multi-turn, agent loop with 4 tools, both gemini-2.5-flash and gemini-2.5-pro, result correctness validation, non-Gemini models unaffected)


YouNeedCryDear:
@paxiaatucsdedu Does this error still occur after your fix https://github.com/oracle/langchain-oracle/pull/108

fede-kamel:
@paxiaatucsdedu Correct — this is a different issue from #108.

**#108** fixed **streaming aggregation**: how tool call chunks get assembled into complete tool calls (fragmented IDs for GPT-OSS, parallel index collisions for Grok). That's about parsing the *inbound* streaming response from the model.

**#111** fixes **outbound message structure**: when Gemini makes N parallel tool calls in one turn, the *conversation history sent back* to the model must have 1:1 function-call-to-function-response pairing per turn. One AIMessage with N tool_calls + N ToolMessages gets rejected by the Gemini API with:

> "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn."

The fix flattens that into N sequential (AIMessage, ToolMessage) pairs before sending.

The issue still persists on latest main (after #108 merge). I confirmed by reproducing against the current main — Gemini parallel tool calls still fail with the 400 INVALID_ARGUMENT error. #108 doesn't touch the message formatting path that this PR fixes.

fede-kamel:
@paxiaatucsdedu Also — both this PR and #109 modify `providers/generic.py`, so there will be merge conflicts. Once #109 is merged, let me know and I'll rebase #111 on top of it to resolve the conflicts.

libs/oci/langchain_oci/chat_models/providers/generic.py
```diff
@@ -230,6 +230,67 @@ def get_role(self, message: BaseMessage) -> str:
             return "TOOL"
         raise ValueError(f"Unknown message type: {type(message)}")
 
+    @staticmethod
```
paxiaatucsdedu: Is it better to move this logic to utils?

PR #116, opened 2026-02-02
Fix Gemini output truncation and add unit + integration tests

fede-kamel:
Gemini models on OCI return long text responses split across multiple content parts in a single API response. GenericProvider.chat_response_to_text was only reading content[0], silently dropping everything after the first part. This caused text output to appear randomly truncated depending on how the model chunked its response.

Root cause in providers/generic.py:

  # before - only first part
  content = message.content[0] if message.content else None
  return content.text if content else ""

  # after - all parts joined
  return "".join(
      part.text for part in message.content
      if getattr(part, "text", None)
  )

Same fix applied to the streaming path (chat_stream_to_text).

Also added auto-normalization of max_output_tokens -> max_tokens for Gemini models. The Google Gemini SDK uses max_output_tokens but the OCI API expects max_tokens. This is now mapped automatically with a UserWarning.

Changes:
- providers/generic.py: concatenate all text content parts instead of content[0], both invoke and stream
- oci_generative_ai.py: max_output_tokens -> max_tokens normalization for google.gemini* models
- Unit tests: 3 new tests covering multipart invoke, multipart stream, and parameter normalization
- Integration test: opt-in diagnostic that sends long-output requests to Gemini and checks for truncation

Note: this fixes client-side truncation (dropped content parts). Long outputs that hit the model's output token cap (finish_reason=max_tokens) are a server-side limit and require pipeline-level chunking.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -50,6 +50,7 @@ from pydantic import BaseModel, ConfigDict, SecretStr, model_validator
 
 from langchain_oci.chat_models.providers import (
     CohereProvider,
+    GeminiProvider,
     GenericProvider,
     MetaProvider,
     Provider,
@@ -170,7 +171,9 @@ class ChatOCIGenAI(BaseChatModel, OCIGenAIBase):
         """Mapping from provider name to provider instance."""
         return {
             "cohere": CohereProvider(),
+            "google": GeminiProvider(),
             "meta": MetaProvider(),
+            "openai": GenericProvider(),
             "generic": GenericProvider(),
         }
 
@@ -218,11 +221,16 @@ class ChatOCIGenAI(BaseChatModel, OCIGenAIBase):
         if stop is not None:
             _model_kwargs[self._provider.stop_sequence_key] = stop
 
+        chat_params = {**_model_kwargs, **kwargs, **oci_params}
+
+        # Apply provider-specific parameter transformations
+        chat_params = self._provider.normalize_params(chat_params)
+
         # Warn if using max_tokens with OpenAI models
         if (
             self.model_id
             and self.model_id.startswith("openai.")
-            and "max_tokens" in _model_kwargs
+            and "max_tokens" in chat_params
         ):
             import warnings
 
@@ -232,8 +240,6 @@ class ChatOCIGenAI(BaseChatModel, OCIGenAIBase):
                 UserWarning,
                 stacklevel=2,
             )
-
-        chat_params = {**_model_kwargs, **kwargs, **oci_params}
 
         if not self.model_id:
             raise ValueError("Model ID is required for chat.")
--- a/libs/oci/langchain_oci/chat_models/providers/base.py
+++ b/libs/oci/langchain_oci/chat_models/providers/base.py
@@ -15,6 +15,20 @@ from pydantic import BaseModel
 class Provider(ABC):
     """Abstract base class for OCI Generative AI providers."""
 
+    @abstractmethod
+    def normalize_params(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Normalize parameters for this provider.
+
+        Subclasses should implement provider-specific parameter transformations.
+
+        Args:
+            params: Dictionary of parameters to normalize
+
+        Returns:
+            Normalized parameters dictionary
+        """
+        ...
+
     @property
     @abstractmethod
     def stop_sequence_key(self) -> str:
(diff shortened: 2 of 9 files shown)


fede-kamel:
Rebased on latest main (includes #109 merge and version bump to 0.2.3). Single clean commit, all CI jobs green.

What the tests prove:

1. Multipart content concatenation (unit tests):
   - Mock a Gemini response with 3 content parts -> invoke returns all 3 joined. Before this fix it only returned the first.
   - Same test for streaming path.
   - These directly reproduce the truncation bug and verify the fix.

2. max_output_tokens normalization (unit test):
   - Passing max_output_tokens to a Gemini model auto-maps it to max_tokens with a UserWarning.
   - Passing max_tokens works as before.
   - Non-Gemini models reject max_output_tokens (expected behavior).

3. Gemini end-to-end (integration tests against real OCI API, us-chicago-1):
   - test_simple_chain PASSED
   - test_chain_with_history PASSED
   - test_chain_batch PASSED
   - test_chain_async PASSED
   - test_stream_chain PASSED
   - test_astream PASSED
   - test_tool_calling_chain PASSED
   - test_enum_constraints_respected PASSED
   - test_numeric_range_constraints PASSED
   - test_schema_sent_to_oci_has_constraints PASSED
   10/10 passed.

4. Real PDF OCR reproduction (google.gemini-2.5-pro, 2-page PDF with 58 numbered lines per page):
   - max_output_tokens=8000 -> 58/58 lines returned, finish_reason=stop
   - max_tokens=8000 -> 58/58 lines returned, finish_reason=stop
   - No truncation in either case.

Unit tests: 128 passed, 0 failures.

fede-kamel:
**Integration tests passed with real Gemini model in Chicago region** ✅

```
Profile: API_KEY_AUTH
Region: us-chicago-1
Model: google.gemini-2.5-flash

test_gemini_basic_invoke PASSED
test_gemini_max_output_tokens_mapping PASSED
test_gemini_both_tokens_params PASSED
test_gemini_streaming PASSED
test_gemini_streaming_with_max_output_tokens PASSED

5 passed in 5.52s
```

The `GeminiProvider` correctly:
- Handles basic invoke/streaming
- Maps `max_output_tokens` → `max_tokens` with appropriate warning
- Prefers `max_tokens` when both params provided

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -170,6 +171,7 @@ def _provider_map(self) -> Mapping[str, Provider]:
         """Mapping from provider name to provider instance."""
         return {
             "cohere": CohereProvider(),
+            "google": GeminiProvider(),
```
paxiaatucsdedu: I think we have docstring, comments, warning log, error log talking about providers across this repo. Would we make sure to include GeminiProvider() in these places?

libs/oci/langchain_oci/chat_models/providers/base.py
```diff
@@ -15,6 +15,31 @@
 class Provider(ABC):
     """Abstract base class for OCI Generative AI providers."""
 
+    # Parameter transformations: maps input param names to provider-specific names
+    # e.g., {"max_output_tokens": "max_tokens"} for Gemini
+    param_transforms: Dict[str, str] = {}
+
+    def normalize_params(self, params: Dict[str, Any]) -> Dict[str, Any]:
```
YouNeedCryDear: This Provider class is just an interface. The implementation probably should be moved to GenericProvider.

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -217,23 +219,11 @@ def _prepare_request(
         if stop is not None:
             _model_kwargs[self._provider.stop_sequence_key] = stop
 
-        # Warn if using max_tokens with OpenAI models
```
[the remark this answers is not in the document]
fede-kamel: Fixed - GenericProvider still works with OpenAI models. The warning is now inline in `_prepare_request` (same location as before this PR). No breaking changes.

libs/oci/langchain_oci/chat_models/providers/base.py
fede-kamel: Done - `Provider` is now a pure interface. `normalize_params` is abstract, and the default implementation is in `GenericProvider`.

libs/js/langchain-oracledb/scripts/move-cjs-to-dist.js
YouNeedCryDear: What is this js file change about?

fede-kamel: Removed - this was an unintended artifact from rebasing on main. The file now matches origin/main.

Issue #103, opened 2026-01-26
Tool Parameter Types are not fully passed to LLM


nrauschcom:
We are currently trying to switch from langchain-google to langchain-oracle in order to use Gemini on OCI.
However, when using a tool-calling agent, we face an issue that the parameter types passed to the LLM are not comprehensive - in our MCP server, we defined a tool `create_files` with the following `args_schema`:
```
{
    "type": "object",
    "properties": {
        "inputs": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "names": {
                        "type": "object",
                        "description": "A raw json object containing ISO 639-1 Language Codes as keys and the respective names of the file in all languages of the parent folder, e.g. {"de": "Dateiname", "en": "File Name"}."
                    },
                    "parentId": {
                        "type": "integer",
                        "format": "int64",
                        "description": "The ID of the folder where the file should be created."
                    },
                    "fileTypeId": {
                        "type": "integer",
                        "format": "int32",
                        "description": "The ID of the file type. Must be a valid File Type for this Folder."
                    }
                },
                "required": [
                    "names",
                    "parentId",
                    "fileTypeId"
                ]
            },
            "description": "Information about the files to be generated."
        }
    }
```

Using `gemini-2.5-flash` on `langchain-google`, the generated tool calls contain all those attributes with logical values. Using `langchain-oracle` however, it seems like only "List" is passed to the LLM and the inner structure of the items is guessed by the LLM.

Is this not a supported feature of `langchain-oracle` right now or are we facing a bug? If it is not currently supported, are there plans on adding this functionality?

PR #118, opened 2026-02-02
Fix nested tool schema and anyOf resolution for Gemini on OCI

paxiaatucsdedu:
## Problem

`GenericProvider.convert_to_oci_tool()` was stripping comprehensive parameter types from tool definitions, causing two critical issues:

1. **Nested schemas lost**: `List[BaseModel]` schemas were reduced to basic types, losing nested object structures
2. **Optional fields broken**: Pydantic v2's `Optional[T]` generates `anyOf` patterns that became `type: "any"`, causing 400 errors on Gemini models

Related issue: https://github.com/oracle/langchain-oracle/issues/103

### Example

**Tool definition:**
```python
class FileInput(BaseModel):
    name: str = Field(description="File name")
    folder_id: int = Field(description="Folder ID")

class CreateFilesTool(BaseTool):
    files: List[FileInput] = Field(description="Files to create")
```

**Before fix** - LLM received incomplete schema:
```json
{
  "files": {
    "type": "array",
    "description": "Files to create"
  }
}
```
Result: LLM hallucinated field names and used wrong types.

**After fix** - LLM receives complete schema:
```json
{
  "files": {
    "type": "array",
    "description": "Files to create",
    "items": {
      "type": "object",
      "properties": {
        "name": {"type": "string", "description": "File name"},
        "folder_id": {"type": "integer", "description": "Folder ID"}
      },
      "required": ["name", "folder_id"]
    }
  }
}
```
Result: LLM generates correct tool calls with proper field names and types.

## Solution

Modified `GenericProvider.convert_to_oci_tool()` to:

1. **Use `model_json_schema()`** instead of `convert_to_openai_function()` to preserve `json_schema_extra` constraints (enum, format, const)
2. **Resolve `$ref` references** using existing `OCIUtils.resolve_schema_refs()` (OCI API doesn't support JSON Schema references)
4. **Resolve `anyOf` patterns** from Pydantic v2's `Optional[T]` to prevent `type: "any"` errors

### Notebook Validation
- Testing all edge cases from PR #109
- **14/14 scenarios pass**:
  - ✅ `json_schema_extra` constraints (enum, format, const)
  - ✅ Pydantic-native constraints (ge/le, gt/lt, pattern, min/max length)
  - ✅ `anyOf` patterns (Optional fields)
  - ✅ `$ref` resolution (nested models, Python Enums)
  - ✅ Nested object schemas
  - ✅ Array item schemas

### Integration Test
- Verified with live Gemini model on OCI
- Tool calls now include correct field names and types
- No more 400 errors from `type: "any"`

## Related
Addresses the same core issues as PR #109 for `GenericProvider` (used by Gemini, Meta Llama, xAI Grok).

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -378,9 +378,17 @@ class GenericProvider(Provider):
         """
         # Check BaseTool first since it's callable but needs special handling
         if isinstance(tool, BaseTool):
-            # Use convert_to_openai_function to get full schema with $defs
-            as_json_schema_function = convert_to_openai_function(tool)
-            parameters = as_json_schema_function.get("parameters", {})
+            # Use model_json_schema() if available to preserve json_schema_extra
+            # constraints (enum, format, etc.) that convert_to_openai_function strips
+            if tool.args_schema and hasattr(tool.args_schema, "model_json_schema"):
+                schema = tool.args_schema.model_json_schema()
+                # Resolve $ref/$defs — OCI doesn't support JSON Schema references
+                schema = OCIUtils.resolve_schema_refs(schema)
+                parameters = schema
+            else:
+                # Fallback to convert_to_openai_function for tools without args_schema
+                as_json_schema_function = convert_to_openai_function(tool)
+                parameters = as_json_schema_function.get("parameters", {})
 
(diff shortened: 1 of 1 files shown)


fede-kamel:
To be clear — I lean towards merging #109 since it covers the Cohere side too and has the allowlist safety net, but this is your call. No issues with either approach, the GenericProvider fix here is solid.

PR #147, opened 2026-02-26
Refactor: get signer/config from sync client's base_client

fede-kamel:
## Summary

Addresses review feedback from PR #124, improving async support with proper thread-safety and cleaner architecture.

## Why Async Support Matters

Without native async, LangChain falls back to `asyncio.to_thread()` which wraps sync calls in a thread pool. This works but has overhead and doesn't scale well.

With native async:
- Multiple LLM calls run concurrently instead of sequentially (10 requests that take 30s one-by-one finish in ~3s)
- In FastAPI/aiohttp servers, `await llm.ainvoke()` doesn't block the event loop
- Streaming with `async for chunk in llm.astream()` delivers tokens without thread overhead

## Changes

- Added `asyncio.Lock` with double-checked locking to prevent race conditions when multiple coroutines initialize the async client
- Access signer/config directly from `self.client.base_client` instead of storing redundant attributes
- `OCIAsyncClient` now implements `__aenter__`/`__aexit__` for proper resource cleanup
- Added helper to detect API version from OCI SDK before falling back to known version
- Using LangChain's `UsageMetadata` directly instead of custom class
- Fixed async integration tests (env var consistency, updated default model)

## Tested Models

| Model | Tests | Time |
|-------|-------|------|
| `meta.llama-3.3-70b-instruct` | 8/8 ✓ | ~12s |
| `meta.llama-4-maverick-17b-128e-instruct` | 8/8 ✓ | ~15s |
| `cohere.command-r-plus-08-2024` | 8/8 ✓ | ~14s |
| `google.gemini-2.5-flash` | 8/8 ✓ | ~11s |
| `xai.grok-2-1212` | 8/8 ✓ | ~13s |
| `openai.gpt-4o` | 8/8 ✓ | ~11s |
| `openai.gpt-4o-mini` | 8/8 ✓ | ~10s |
| `openai.gpt-5` | 8/8 ✓ | ~125s |

## SDK Migration

Once the OCI SDK adds native async support, replacing `OCIAsyncClient` is straightforward:

```python
# Current
self._async_client = OCIAsyncClient(
    service_endpoint=self.service_endpoint,
    signer=base_client.signer,
    config=getattr(base_client, "config", {}),
)

# Future (when SDK supports async)
self._async_client = GenerativeAiInferenceAsyncClient(
    config=self.client.base_client.config,
    signer=self.client.base_client.signer,
    service_endpoint=self.service_endpoint,
)
```

The mixin architecture keeps async logic isolated, so migration is a swap.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/async_mixin.py
+++ b/libs/oci/langchain_oci/chat_models/async_mixin.py
@@ -7,9 +7,9 @@ This module provides async support for ChatOCIGenAI through a mixin class,
 keeping the main module clean and focused.
 """
 
-import asyncio
 import json
 import uuid
+from functools import cached_property
 from typing import Any, AsyncIterator, Dict, List, Optional
 
 from langchain_core.callbacks import AsyncCallbackManagerForLLMRun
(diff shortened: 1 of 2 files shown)


fede-kamel:
## Usage Example

```python
from langchain_oci import ChatOCIGenAI
import asyncio

llm = ChatOCIGenAI(
    model_id="meta.llama-3.3-70b-instruct",
    compartment_id="ocid1.compartment...",
    service_endpoint="https://inference.generativeai.us-chicago-1.oci.oraclecloud.com",
    auth_type="API_KEY",
)

async def main():
    # Single async call
    response = await llm.ainvoke("What is the capital of France?")
    print(response.content)

    # Async streaming
    async for chunk in llm.astream("Write a haiku about clouds"):
        print(chunk.content, end="", flush=True)

    # Concurrent requests (this is where async shines)
    questions = [
        "What is 2+2?",
        "What is the capital of Argentina?",
        "Who wrote Hamlet?",
    ]
    responses = await asyncio.gather(*[llm.ainvoke(q) for q in questions])
    for q, r in zip(questions, responses):
        print(f"{q} -> {r.content}")

asyncio.run(main())
```

YouNeedCryDear:
I think the get async client and lock implementation is still no optimal, which might introduce race condition. Could you please refer to what langchain-aws is doing? either use threading.lock or use pydantic post init? @fede-kamel

PR #148, opened 2026-02-26
feat: Add agent shared config and refactor react agent to module

fede-kamel:
## Summary

Introduce shared agent infrastructure as the foundation for the deep research agent stack.

- Add `AgentConfig` base class and `_build_llm()` shared builder in `agents/common.py` — reusable config for all OCI agent types
- Refactor react agent from single file (`react.py`) to module (`react/`) using the shared config
- Add `_compat.py` with langgraph schema compatibility patches for upstream `OmitFromSchema`/`NotRequired` annotation bugs
- Fix tool schema conversion to use `tool_call_schema` instead of `args_schema` for deepagents middleware compatibility
- Add `supports_tool_choice` to GenericProvider for structured output
- Fix multi-turn tool call counting to scope to current turn only
- Add configurable request timeout via `auth.py`

## What Changed

| File | Change |
|------|--------|
| `agents/common.py` | New — `AgentConfig`, `_build_llm`, `_get_agent_factory`, `_filter_none` |
| `agents/react.py` | Deleted — replaced by module |
| `agents/react/__init__.py` | New — re-exports `create_oci_agent` |
| `agents/react/agent.py` | New — `create_oci_agent` refactored to use `AgentConfig` |
| `agents/__init__.py` | Updated — exports `AgentConfig` |
| `_compat.py` | New — langgraph schema patches |
| `chat_models/providers/base.py` | New — `_overlay_schema_extras`, `supports_tool_choice` |
| `chat_models/providers/cohere.py` | Fix — use `tool_call_schema` for tool conversion |
| `chat_models/providers/generic.py` | Fix — `_get_tool_parameters`, turn-scoped tool counting |
| `chat_models/oci_generative_ai.py` | Fix — `tool_choice="required"` for structured output |
| `common/auth.py` | New — configurable timeout via `_resolve_timeout` |

## Test plan

- [x] 16 unit tests passing (`test_react.py`)
- [x] 45 unit tests passing (`test_tool_schema_constraints`, `test_tool_call_turn_scoping`)
- [x] Integration tests passing (`test_react_integration`, `test_tool_calling`, `test_response_format`)
- [x] 60 integration tests passing against live ADB + OpenSearch (full stack)

## Context

This is **PR 1 of 13** in the deep research agent stack (includes structured output fix from #190). See the comment below for the full breakdown.


fede-kamel:
Deep Research (deepsearch) integration tests (scoped)

Command:
```bash
cd libs/oci
OCI_AUTH_TYPE=API_KEY OCI_AUTH_PROFILE=API_KEY_AUTH OCI_SERVICE_ENDPOINT=https://inference.generativeai.us-chicago-1.oci.oraclecloud.com OCI_COMPARTMENT_ID=ocid1.tenancy.oc1..aaaaaaaah7ixt2oanvvualoahejm63r66c3pse5u4nd4gzviax7eeeqhrysq OCI_EMBEDDING_MODEL_ID=ocid1.generativeaimodel.oc1.us-chicago-1.amaaaaaask7dceya6a7ddhqkyimvl3olp3wmllmbedfsghe633fn2e63kicq OPENSEARCH_ENDPOINT=https://ai-dev.observ.us-ashburn-1.ocs.oraclecloud.com:9200 OPENSEARCH_USERNAME=ai_user OPENSEARCH_PASSWORD='***' OPENSEARCH_INDEX=observai_development_docs_agebre OPENSEARCH_VECTOR_FIELD=vector_field ADB_DSN='tcps://adb.ca-toronto-1.oraclecloud.com:1522/g4549e8afff78c6_deepresearch_low.adb.oraclecloud.com' ADB_USER=ADMIN ADB_PASSWORD='***' ADB_TABLE_NAME=VECTOR_DOCUMENTS_LC ADB_TEST_KEYWORD=alignment poetry run pytest -q tests/integration_tests/agents -k 'deep_research or deepsearch'
```

Result: `11 passed, 34 deselected` (4m11s). No aiohttp unclosed session/connector warnings after adding explicit LLM cleanup.\n
LLM used by tests: `google.gemini-2.5-pro`. Embeddings configured via `OCI_EMBEDDING_MODEL_ID` (Cohere embed model).

aojah1:
This PR is very important from an Agentic AI implementation point of view, and in particular for some of the projects I am working with data sources like EBS and PeopleSoft. I have tested all three natives - Object Storage, ADB, and WebSearch worked for me as expected. I especially liked the object storage feature as I needed it for my use case. I tested this gist entirely, and the results are as expected - https://gist.github.com/fede-kamel/3a5f36c02eb591c1366bfaa16df35cf8

fede-kamel:
@YouNeedCryDear could you take another look at this? I refactored the deep-research datastore internals to hang off standard LangChain `VectorStore` / optional `BaseRetriever` contracts while keeping the public OCI API unchanged, and I reran the relevant unit/live integration validation on this branch.

YouNeedCryDear (changes requested):
@fede-kamel I think there are a lot of changes that were not supposed to be here. Please pull the latest main first.

YouNeedCryDear:
@fede-kamel Also for the changes made in langchain-oracledb, I think you need another reviewer from @fileames team as well.

fede-kamel:
## Final update (2026-03-19)

@YouNeedCryDear Sorry about the earlier state of the branch -- the deep research work was developed before PR #147 landed and I pushed without rebasing first, which caused all those unintended reverts. That is fixed now. The branch is clean against main and the diff only contains deep-research-related changes.

### All review feedback addressed

**Arthur's 3 inline comments** -- all resolved:
- `oci_generative_ai.py` -- removed `oci_signer`/`oci_config` from base class
- `async_support.py` -- restored dynamic API version detection and `__aenter__`/`__aexit__`
- 6 files that reverted PR #147 have been reset to match main exactly

**Luigi's 2 compatibility issues** -- both fixed:
- Tool schema conversion now uses `tool_call_schema` (not `args_schema`), fixing `PydanticInvalidForJsonSchema` with deepagents middleware
- `_compat.py` patches langgraph `create_model` for `OmitFromSchema`/`NotRequired` handling
- Luigi's exact hello world scenario (`test_oci01.py`) runs end-to-end

### Additional work

- Runtime prerequisite checks: clear errors for Python < 3.11 or missing deepagents
- Env var consistency: `OCI_AUTH_PROFILE` used everywhere (was `OCI_CONFIG_PROFILE` in places)
- Integration tests read auth from environment instead of hardcoding profiles

### Test results (free tier, API_KEY auth)

```
Unit tests (Python 3.9)            309 passed, 33 skipped
Unit tests (Python 3.12)           106 passed
Integration tests                   52 passed, 0 skipped
  - test_adb.py                      5 passed
  - test_opensearch.py              10 passed
  - test_agents.py                   6 passed
  - test_deep_research_integration  14 passed
  - test_react_integration          14 passed
  - test_sample_scripts              2 passed
  - test_compile                     1 passed
Luigi compat validation            7/7 passed
Datastore validation (ADB+OS)      7/7 passed
```

Live datastores validated:
- ADB: `deepresearch_low` (free tier, ca-toronto-1)
- OpenSearch: `observai_diagnostic-patterns` (dev, us-ashburn-1)
- Models: Gemini 2.5 Flash/Pro, Cohere embed-v4.0

### Generated research outputs

- ADB report: 9,211 words (Gemini 2.5 Flash, 65k max tokens)
- OpenSearch report: 6,179 words (Gemini 2.5 Pro, 65k max tokens)

### Gists

- Samples + Luigi compat + outputs: https://gist.github.com/fede-kamel/92b9a515155f4332f794d491602b6e79
- PR 148 validation status (v10): https://gist.github.com/fede-kamel/00cb568227912735e9717ddc12b649c4

### Remaining

- `libs/oracledb` changes need a reviewer from @fileames team (as Arthur noted)

fileames (dismissed):
Thanks for the updates!

libs/oci/langchain_oci/agents/datastores/__init__.py
YouNeedCryDear: I don't think we need the datastores folder. We can move the tools and vectorstores to the top level. Those should be generic to not only agent use cases.

libs/oci/langchain_oci/agents/deep_research/agent.py
YouNeedCryDear: I think there are several refactor we can do. There are certainly duplication in both react agent and deep research agent. Can we consolidate those into an abstraction and reuse code as much as possible?

libs/oci/langchain_oci/agents/deep_research/agent.py
```diff
@@ -0,0 +1,335 @@
...
+    tools: Optional[Sequence[Union[BaseTool, Callable[..., Any]]]] = None,
+    *,
+    # Datastores - if provided, auto-routing search is enabled
+    datastores: Optional[Dict[str, VectorDataStore]] = None,
+    default_datastore: Optional[str] = None,
+    default_store: Optional[str] = None,  # Alias for default_datastore
+    embedding_model: Any = None,
+    top_k: int = 5,
+    # OCI-specific options
+    model_id: str = "google.gemini-2.5-pro",
```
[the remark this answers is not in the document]
fede-kamel: Both `create_oci_agent` and `create_deep_research_agent` return `CompiledStateGraph`, which is already a LangChain `Runnable` — supports `.invoke()`, `.ainvoke()`, `.stream()`, `.astream()`, and pipes. The config models (`AgentConfig`, `DeepResearchConfig`) handle all the setup; the returned object is a clean Runnable ready to use.

fede-kamel:
PR is ready for review.

cc: @YouNeedCryDear

fede-kamel:
@YouNeedCryDear — Per your feedback on PR size, I've split this into a stack of smaller, focused PRs (~500 lines each). Here's the breakdown:

**Independent fixes (can merge in any order):**
- #190 — Structured output + tool calling fixes (~500 lines)
- #191 — OracleDB document chunking (~296 lines)

**This PR (#148) is now trimmed to just the foundation:**
- Agent shared config (`AgentConfig`, `_build_llm`) + react agent module refactor (~475 lines)

**Remaining stack (merge in order after #148):**
- #192 — Datastore vectorstore base + ADB adapter
- #193 — OpenSearch vectorstore adapter
- #194 — Datastore tool framework
- #195 — Datastore factory + package wiring
- #196 — Datastore unit tests
- #197 — Deep research agent
- #198 — Deep research unit tests
- #199 — Agent integration test infra
- #200 — Datastore integration tests
- #201 — Deep research integration tests
- #202 — ADB sample + tutorial
- #203 — OpenSearch sample + docs

All 76 unit tests and 58 integration tests pass against live ADB and OpenSearch. Each PR in the stack includes all predecessor code so CI can run independently.

Happy to start reviews from this PR whenever you're ready.

libs/oci/langchain_oci/_compat.py
```diff
@@ -0,0 +1,114 @@
...
+    return cast(
+        type[BaseModel],
+        create_model(model_name, **fields),  # type: ignore[call-overload]
+    )
+
+
+def _needs_patch() -> bool:
+    """Check if the current langgraph version needs the schema patch."""
+    try:
+        from langgraph._internal import _pydantic as langgraph_pydantic
```
YouNeedCryDear: Is it because of the old langchain version we need to support or it is just different API for different langgraph version?

libs/oci/langchain_oci/_compat.py
```diff
@@ -0,0 +1,114 @@
...
+            )
+
+    patched_create_model._langchain_oci_patched = True  # type: ignore[attr-defined]
+    langgraph_pydantic.create_model = patched_create_model
+
+    # Also patch modules that imported create_model directly before us,
+    # since `from X import create_model` copies the reference.
+    import sys
+
+    for mod_name in ("langgraph.pregel.main", "langgraph.pregel"):
```
YouNeedCryDear: This doesn't feel like a safe way. Is it possible to condition on the langgraph version and only create alias for the create_model function?

libs/oci/langchain_oci/_compat.py
```diff
@@ -0,0 +1,114 @@
...
+        field_definitions: Optional[Dict[str, Any]] = None,
+        root: Optional[Any] = None,
+    ) -> type[BaseModel]:
+        try:
+            return original_create_model(
+                model_name,
+                field_definitions=field_definitions,
+                root=root,
+            )
+        except Exception as ex:
```
YouNeedCryDear: can we narrow down the exception here?

libs/oci/langchain_oci/agents/common.py
```diff
@@ -0,0 +1,125 @@
...
+        if isinstance(config.auth_type, OCIAuthType)
+        else config.auth_type
+    )
+
+    merged_kwargs = {**config.model_kwargs}
+    if config.temperature is not None:
+        merged_kwargs["temperature"] = config.temperature
+    if config.max_tokens is not None:
+        if config.model_id and config.model_id.startswith("openai."):
+            merged_kwargs["max_completion_tokens"] = config.max_tokens
```
YouNeedCryDear: Do you think it is better to change it on the provider level?

libs/oci/langchain_oci/chat_models/providers/base.py
```diff
@@ -124,6 +124,34 @@ def process_stream_tool_calls(
         """Process streaming tool calls from event data into chunks."""
         ...
 
+    @staticmethod
+    def _overlay_schema_extras(
```
YouNeedCryDear: probably better be a function in util?

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
```diff
@@ -393,7 +393,12 @@ def with_structured_output(
         if method == "function_calling":
             if schema is None:
                 raise ValueError("Schema must be provided for function_calling method.")
-            llm = self.bind_tools([schema], **kwargs)
+            # Force tool use when the provider supports tool_choice, so the
+            # model returns structured output instead of free-form text.
+            bind_kwargs: Dict[str, Any] = {**kwargs}
+            if self._provider.supports_tool_choice:
+                bind_kwargs["tool_choice"] = "required"
```
YouNeedCryDear, line 389: Have this been validated to models other than openai?

libs/oci/langchain_oci/common/auth.py
```diff
@@ -3,8 +3,15 @@
...
+import os
 from enum import Enum
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple, Union
+
+# Default timeout: (connect_timeout, read_timeout) in seconds.
+# Override via OCI_REQUEST_TIMEOUT env var (read timeout only)
+# or pass timeout= to create_oci_client_kwargs.
+_DEFAULT_CONNECT_TIMEOUT = 10
+_DEFAULT_READ_TIMEOUT = 240
```
YouNeedCryDear, line 16: I think there is default timeout defined in for both OCI GenAI API call and general OCI API call.  Any reason  not using those?

libs/oci/langchain_oci/chat_models/providers/cohere.py
```diff
@@ -612,17 +612,33 @@ def convert_to_oci_tool(
...
+                # Overlay json_schema_extra constraints from args_schema
+                if tool.args_schema:
+                    self._overlay_schema_extras(schema, tool.args_schema)
                 # Resolve $ref/$defs and anyOf — OCI doesn't support them
                 schema = OCIUtils.resolve_schema_refs(schema)
                 schema = OCIUtils.resolve_anyof(schema)
                 schema = OCIUtils.sanitize_schema(schema)
                 properties = schema.get("properties", {})
+            if properties is None:
+                as_json_schema_function = convert_to_openai_function(tool)
```
YouNeedCryDear, line 632: Is there any difference between how cohere expect the function vs openai?

libs/oci/langchain_oci/_compat.py
fede-kamel: Neither really — it's a workaround for an upstream pydantic bug where `OmitFromSchema + NotRequired` (used by deepagents' tool schemas) makes langgraph's `create_model` blow up. So the patch isn't tied to supporting older langchain, it's compensating for a specific pydantic/langgraph interaction that deepagents schemas happen to hit. I added an upper-bound version check so the patch will auto-retire once upstream fixes it, rather than applying forever.

libs/oci/langchain_oci/_compat.py
fede-kamel: Fair point, it was way too silent. Gated it on a langgraph version check now, and if a module that imported `create_model` has a reference that's drifted from what we're patching, it'll log a warning instead of just quietly skipping. That way if langgraph moves stuff around internally we'll actually notice rather than have the patch silently become a no-op.

libs/oci/langchain_oci/_compat.py
fede-kamel: Good call, the class-name string match was pretty janky. Switched it to import `PydanticForbiddenQualifier` directly (with a fallback to `TypeError` for older pydantic versions that don't expose it at the top level), and the except clause now catches those types properly.

libs/oci/langchain_oci/agents/common.py
fede-kamel: Yeah, you're right — that branch had no business being in the agent layer. Moved it to a new `OpenAIProvider` subclass that does the remap in `normalize_params`, same pattern `GeminiProvider` already uses for `max_output_tokens → max_tokens`. `agents/common.py` just passes `max_tokens` through regardless of provider now. Also dropped the ad-hoc warning in `_prepare_request` since it's redundant with the silent normalize.

libs/oci/langchain_oci/chat_models/providers/base.py
fede-kamel: Yeah, it belongs next to `resolve_schema_refs` and `sanitize_schema` in `OCIUtils`. I left it as a static method on `Provider` in this PR just to keep the diff focused on the bigger items you flagged — happy to do the move as a quick follow-up if you'd rather not bundle it in here.

libs/oci/langchain_oci/chat_models/oci_generative_ai.py
fede-kamel: Good flag — I ran the full matrix to be sure this doesn't regress anyone. OpenAI (gpt-4.1, gpt-5.1 / 5.2 / 5.4 + mini + nano), Meta Llama (3.3-70b, 4-scout, 4-maverick), Gemini 2.5 (flash, flash-lite, pro), Grok (grok-4-fast, grok-3-mini), and Cohere (command-a, command-r-plus) — all pass across `test_structured_output_no_docstring`, `_with_enum`, and `_nested`.

I also added a small escape hatch: if a caller passes an explicit `tool_choice` in kwargs, we honor it instead of overriding to `"required"`. Just in case a future model handles `"required"` poorly and someone wants to opt out without having to work around the library.

libs/oci/langchain_oci/common/auth.py
fede-kamel: The SDK default is (10, 60) — the 240s read timeout came from hitting actual timeouts on long reasoning responses (gpt-5 / o3-style) and streaming completions that legitimately take over a minute. Probably should've had a comment explaining that from the start; I added one now. Users can still dial it back to the SDK default via `OCI_REQUEST_TIMEOUT` or the `timeout=` kwarg if they want tighter bounds.

libs/oci/langchain_oci/chat_models/providers/cohere.py
fede-kamel: No functional difference where it actually matters — the fallback only reads `parameters.properties` and `parameters.required`, which are plain JSON Schema and the same across both providers. The OCI-specific Cohere shape gets built right below via `self.oci_tool_param(...)`, so whatever format `convert_to_openai_function` hands us, we only extract the JSON-Schema subset out of it.

That said, you're right that the variable name `as_json_schema_function = convert_to_openai_function(tool)` reads like we're forwarding OpenAI format straight to Cohere, which isn't what's happening. Happy to rename it to `fallback_tool_schema` or similar for clarity as a quick follow-up.

fede-kamel:
@YouNeedCryDear I pushed 9924f83 + 227734b addressing your comments — CI's green. Quick rundown of what changed:

- `_compat.py` — version-gated the langgraph monkey patch so it auto-retires once the upstream pydantic bug gets fixed, and it now logs a warning if the `sys.modules` rebind target has drifted (previously it silently did nothing). Also switched the `except Exception` + class-name match to a real `except (PydanticForbiddenQualifier, TypeError)`.
- `providers/generic.py` + friends — added an `OpenAIProvider(GenericProvider)` that does the `max_tokens → max_completion_tokens` remap in `normalize_params`, same pattern as `GeminiProvider`. The `openai.`-prefix branch is gone from `agents/common.py` and the ad-hoc warning is gone from `_prepare_request`.
- `oci_generative_ai.py` — `with_structured_output` now honors a caller-supplied `tool_choice` in kwargs rather than always overriding with `"required"`, so there's an escape hatch if a model handles it poorly.
- `common/auth.py` — documented the 240s read-timeout override (the SDK default is 60s, which isn't enough for long reasoning / streaming).

Replied inline on the other four — the `_overlay_schema_extras` move and the cohere variable rename I left as small follow-ups to keep this PR focused.

On the tool_choice validation question I ran the full structured-output matrix across providers:

| Family | Models | Result |
|---|---|---|
| OpenAI | gpt-4.1, gpt-5.1, gpt-5.2, gpt-5.4, gpt-5.4-mini, gpt-5.4-nano, gpt-oss-20b, gpt-oss-120b | ✓ |
| Meta Llama | 3.3-70b, 4-scout-17b, 4-maverick-17b | ✓ |
| Gemini | 2.5-flash, 2.5-flash-lite, 2.5-pro | ✓ |
| Grok | grok-4-1-fast-non-reasoning, grok-3-mini-fast | ✓ |
| Cohere | command-a-03-2025, command-r-plus-08-2024 | ✓ |

287 unit tests + 167 integration tests passing, ruff + mypy clean, no regressions.

libs/oci/langchain_oci/chat_models/providers/cohere.py
fede-kamel: Went and actually verified it rather than just reasoning about it. Ran the same tool through both paths:

```python
@tool
def weather(city: str, units: str = "fahrenheit") -> str:
    """Get weather. Args: city: City name. units: Temperature units."""

# Primary path (cohere.py:621-630):
primary = weather.tool_call_schema.model_json_schema()
primary = OCIUtils.resolve_schema_refs(primary)
primary = OCIUtils.resolve_anyof(primary)
primary = OCIUtils.sanitize_schema(primary)
# -> properties: {'city': {'type':'string'}, 'units': {'type':'string', 'default':'fahrenheit'}}
# -> required:   {'city'}

# Fallback path (cohere.py:631-635):
fallback = convert_to_openai_function(weather)['parameters']
# -> properties: {'city': {'type':'string'}, 'units': {'type':'string', 'default':'fahrenheit'}}
# -> required:   ['city']
```

What `oci_tool_param(...)` actually consumes from that — the `{name, type, is_required}` tuple for each parameter — is identical between the two paths:

```
primary:  {'city': {'type': 'string', 'required': True}, 'units': {'type': 'string', 'required': False}}
fallback: {'city': {'type': 'string', 'required': True}, 'units': {'type': 'string', 'required': False}}
```

Also did an end-to-end call against `cohere.command-a-03-2025` to be thorough:
```
tool_calls: [{'name': 'get_weather', 'args': {'city': 'Paris', 'units': 'metric'}, ...}]
```
Cohere picked up the tool correctly.

One thing worth mentioning: the fallback almost never fires in practice. Both `@tool`-decorated functions (which become `StructuredTool`) and plain `Tool` instances come out of langchain_core with `tool_call_schema` already set and having `model_json_schema`, so the fallback is really defensive code for legacy / custom `BaseTool` subclasses that leave it unset. But even in that edge case the two paths converge on the same JSON-Schema subset, so no Cohere/OpenAI format divergence. Still going to rename `as_json_schema_function` in the follow-up for clarity.

libs/oci/langchain_oci/_compat.py
fede-kamel: You were right to push on this — the global import-time patch was the wrong shape. Rewrote it.

`_compat.py` is gone. In its place there's a `_langgraph_schema_fallback` context manager in `agents/common.py`, and the agent factories (`create_oci_agent` here on #148, `create_deep_research_agent` on #197) wrap their `create_agent_func(...)` / `create_deep_agent(...)` call with it:

```python
with _langgraph_schema_fallback():
    return create_agent_func(**agent_kwargs)
```

The patch is only live for the duration of that call and restored cleanly on exit. No side effects on import, nothing to version-gate, nothing to remove from `__init__.py`. If someone uses this package without ever calling one of our agent factories, langgraph internals are never touched. The exception catch is narrowed to real types (`PydanticForbiddenQualifier` imported at module load, with `TypeError` as a fallback).

Also confirmed no behavioral regression — 347 unit tests still green across the stack, and the integration tests I ran against real OCI (Meta Llama, Cohere, Gemini, Grok, OpenAI) still pass.

libs/oci/langchain_oci/chat_models/providers/base.py
fede-kamel: Did it in this PR rather than deferring — moved to `OCIUtils.overlay_schema_extras` alongside `resolve_schema_refs` / `sanitize_schema`. Callers in `GenericProvider` and `CohereProvider` now go through `OCIUtils.overlay_schema_extras(...)`.

YouNeedCryDear (approved):

PR #159, opened 2026-03-13
bugfix(oci): handle GPT-5 parameter compatibility

### LINKED ISSUE #214: ChatOCIGenAI: 400 errors on GPT-5 / OpenAI models due to unsupported temperature / top_p / max_tokens
## Bug

`ChatOCIGenAI` sends `temperature`, `top_p`, and `max_tokens` on every request. OCI rejects some of these parameters for OpenAI / GPT-5 models at request time, so users get unrecoverable HTTP 400s on what is otherwise a clean call.

Observed:

- `temperature` / `top_p` with non-default values against **GPT-5 legacy** → `400 unsupported_value`
- `max_tokens` against **OpenAI models** → `400 unsupported_parameter` (the model expects `max_completion_tokens` instead)

Repro (against live OCI GenAI):

```python
from langchain_oci import ChatOCIGenAI

chat = ChatOCIGenAI(
    model_id=\"openai.gpt-5\",
    service_endpoint=\"...\",
    compartment_id=\"...\",
    model_kwargs={\"temperature\": 0.2, \"max_tokens\": 256},
)
chat.invoke(\"hi\")  # → oci.exceptions.ServiceError 400
```

## Proposed fix

Rather than hardcoding per-model predicates (which won't survive new model releases), catch the OCI `ServiceError` 400, parse the structured error to identify the rejected parameter, fix or rename it on the request, and retry. Concretely:

- `unsupported_value` → drop the offending param (e.g. `temperature`)
- `unsupported_parameter` with a known rename → convert it (`max_tokens` → `max_completion_tokens`)
- `unsupported_parameter` without a rename → drop it (e.g. `top_p`)
- Cap retries at 3 to handle multiple rejected params in one request
- Emit a warning per auto-fixed param so users see what changed

This works for current and future models without model-ID branching.

## Linked PR

PR #159.

fede-kamel:
## Bug
OCI rejects certain parameters for some models at request time:
- `temperature` / `top_p` with non-default values on GPT-5 legacy → `400 unsupported_value`
- `max_tokens` on OpenAI models → `400 unsupported_parameter` (requires `max_completion_tokens`)

Previously, users of `ChatOCIGenAI` with these models would get unrecoverable 400 errors.

## Fix
Instead of hardcoding model-specific predicates, the fix catches OCI `ServiceError` 400s, parses the structured error response to identify the rejected parameter, fixes or renames it on the request object, and retries. This works for any current or future model — no model ID checks needed.

- `unsupported_value` errors → param is removed (e.g. `temperature`)
- `unsupported_parameter` with a known rename → param is converted (e.g. `max_tokens` → `max_completion_tokens`)
- `unsupported_parameter` without rename → param is removed (e.g. `top_p`)
- Up to 3 retries to handle multiple rejected params in one request
- User gets a warning for each auto-fixed param

## Validation
**Unit tests (4 new):**
- `test_retry_strips_unsupported_temperature`
- `test_retry_converts_max_tokens_to_max_completion_tokens`
- `test_retry_strips_unsupported_top_p`
- `test_no_retry_on_unrelated_400`

**Integration test:** `test_openai_gpt5_tool_calling` — passes against live OCI GPT-5, retry visible in warnings

**Lint:** `ruff check` + `ruff format` — pass

Closes #214

### DIFF
--- a/libs/oci/langchain_oci/chat_models/async_mixin.py
+++ b/libs/oci/langchain_oci/chat_models/async_mixin.py
@@ -18,7 +18,11 @@ from langchain_core.messages import AIMessage, AIMessageChunk, BaseMessage, Tool
 from langchain_core.messages.ai import UsageMetadata
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
 
-from langchain_oci.common.async_support import OCIAsyncClient
+from langchain_oci.common.async_support import OCIAsyncClient, OCIAsyncRequestError
+from langchain_oci.common.param_compat import (
+    PARAM_RETRY_ATTEMPTS,
+    adjust_request_for_param_error,
+)
 from langchain_oci.llms.utils import enforce_stop_tokens
 
 
@@ -110,16 +114,30 @@ class ChatOCIGenAIAsyncMixin:
             messages, stop, stream=False, **kwargs
         )
 
-        # Get single response
+        # Get single response, retrying after fixable 400 parameter errors
+        # (mirrors the sync _chat_with_param_retry; e.g. legacy openai.gpt-5
+        # rejects non-default temperature/top_p with 400 unsupported_value).
         response_data = None
-        async for data in client.chat_async(
-            compartment_id=request_data["compartment_id"],
-            chat_request_dict=request_data["chat_request_dict"],
-            serving_mode_dict=request_data["serving_mode_dict"],
-            stream=False,
-        ):
-            response_data = data
-            break
+        for attempt in range(PARAM_RETRY_ATTEMPTS):
+            try:
+                async for data in client.chat_async(
+                    compartment_id=request_data["compartment_id"],
+                    chat_request_dict=request_data["chat_request_dict"],
+                    serving_mode_dict=request_data["serving_mode_dict"],
+                    stream=False,
+                ):
+                    response_data = data
+                    break
+                break
+            except OCIAsyncRequestError as e:
+                if (
+                    e.status != 400
+                    or attempt == PARAM_RETRY_ATTEMPTS - 1
+                    or not adjust_request_for_param_error(
+                        e.body, request_data["chat_request_dict"]
+                    )
+                ):
+                    raise
 
         if response_data is None:
             raise RuntimeError("No response received from OCI GenAI")
@@ -188,12 +206,40 @@ class ChatOCIGenAIAsyncMixin:
         if reset is not None:
             reset()
 
-        async for event_data in client.chat_async(
-            compartment_id=request_data["compartment_id"],
-            chat_request_dict=request_data["chat_request_dict"],
-            serving_mode_dict=request_data["serving_mode_dict"],
-            stream=True,
-        ):
+        async def _events_with_param_retry() -> AsyncIterator[Dict[str, Any]]:
+            """Open the stream, retrying after fixable 400 parameter errors.
+
+            A non-200 surfaces before the first event, so only the first
+            fetch is retried (mirroring the sync _chat_with_param_retry);
+            mid-stream errors propagate unchanged.
+            """
+            for attempt in range(PARAM_RETRY_ATTEMPTS):
+                stream = client.chat_async(
+                    compartment_id=request_data["compartment_id"],
+                    chat_request_dict=request_data["chat_request_dict"],
+                    serving_mode_dict=request_data["serving_mode_dict"],
+                    stream=True,
+                )
+                try:
+                    first = await stream.__anext__()
+                except StopAsyncIteration:
+                    return
+                except OCIAsyncRequestError as e:
+                    if (
+                        e.status != 400
+                        or attempt == PARAM_RETRY_ATTEMPTS - 1
+                        or not adjust_request_for_param_error(
+                            e.body, request_data["chat_request_dict"]
+                        )
+                    ):
+                        raise
+                    continue
+                yield first
+                async for event in stream:
+                    yield event
+                return
+
+        async for event_data in _events_with_param_retry():
             if not self._provider.is_chat_stream_end(event_data):  # type: ignore[attr-defined]
                 # Process streaming content
                 delta = self._provider.chat_stream_to_text(event_data)  # type: ignore[attr-defined]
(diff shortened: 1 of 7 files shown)


fede-kamel:
I force-updated this PR to a clean `main`-based branch.
Current scope is only the GPT-5 compatibility bug:
- legacy OCI `openai.gpt-5` sampling parameter normalization in `ChatOCIGenAI`
- OpenAI agent `max_tokens` -> `max_completion_tokens` mapping
- unit coverage for both paths
- live GPT-5 ReAct integration coverage

Validation on this branch:
- `poetry run pytest tests/unit_tests/chat_models/test_oci_generative_ai.py tests/unit_tests/agents/test_react.py -q` -> `61 passed`
- `poetry run pytest tests/integration_tests/agents/test_react_integration.py::test_openai_gpt5_tool_calling -q` -> `1 passed`
- `poetry run ruff check langchain_oci/chat_models/oci_generative_ai.py langchain_oci/agents/react.py tests/unit_tests/chat_models/test_oci_generative_ai.py tests/unit_tests/agents/test_react.py tests/integration_tests/agents/test_react_integration.py` -> passed

fede-kamel:
Hi @YouNeedCryDear  — this PR has been rebased on latest main and cleaned up. It's now a single commit with no dependency on the deep research PR (#148).

**The issue is confirmed real** — GPT-5 on OCI currently rejects:
1. Non-default `temperature` / `top_p` (`400 unsupported_value`)
2. `max_tokens` (requires `max_completion_tokens` instead)

The fix:
- Strips unsupported sampling params for GPT-5 legacy models with a warning
- Auto-converts `max_tokens` → `max_completion_tokens` for all OpenAI models (previously only warned)
- Keeps newer GPT-5.x models on the normal path

All CI green, unit + integration tests passing locally against real OCI. Would appreciate a review when you get a chance.

fileames:
Are these relevant: 
  - libs/oci/langchain_oci/chat_models/oci_generative_ai.py:584 / :660: the retry path passes only ServiceError.message into _handle_unsupported_param_error. For the OpenAI-style body this helper is written for, {"error": {"param": ..., "code": ...}}, the OCI SDK does not promote that nested payload into .message; it keeps it in ServiceError.args[0]["error"]. That means the retry can miss the structured error, and if .message is None, line 146 can raise TypeError on "maxTokens" in message, masking the original 400. Pass the ServiceError or parse e.args[0]["error"], and make the helper robust to non-string/empty messages.

  - libs/oci/langchain_oci/chat_models/async_mixin.py:115 and :191: the same unsupported-param failure is still unhandled for ainvoke / astream. Async serializes the request and calls OCIAsyncClient.chat_async, which raises RuntimeError on non-200 in libs/oci/langchain_oci/common/async_support.py:241. The new sync-only ServiceError retry will not run there, so GPT-5 agent/ tool-calling still fails for async users with the same params. The retry/param-adjustment should be shared across sync and async paths, or moved before serialization where both paths benefit.

fede-kamel:
@fileames Thanks for the sharp review — I verified both points and you're right on both. Findings:

**1. `.message` / TypeError — confirmed, reproduced.** In the SDK, `ServiceError.__init__` assigns `self.message = message` *before* its `if not message:` fallback (the fallback only feeds the formatted string), so for an OpenAI-style passthrough body with no top-level `message` key, `e.message is None`. Reproduced on this branch:

```text
e.message = None
TypeError raised as predicted: argument of type 'NoneType' is not iterable
```

i.e. the `"maxTokens" in message` check throws and masks the original 400 — and since the structured `{"error": {"param": ..., "code": ...}}` payload is never promoted into `.message`, the structured branch can't fire for exactly the body shape it was written for.

**2. Async gap — confirmed from source.** `common/async_support.py` raises a plain `RuntimeError` on non-200, so the sync-only `ServiceError` retry never runs for `ainvoke`/`astream`.

**One more thing that changed under this PR:** main has since gained `OpenAIProvider.normalize_params`, which proactively renames `max_tokens → max_completion_tokens` at request-build time — before serialization, so it already covers both sync and async. That makes the rename-retry portion of this PR redundant; the still-novel part is the `unsupported_value` handling (non-default `temperature`/`top_p` on GPT-5 legacy).

**Rework plan:** rebase onto current main, drop the redundant rename path, pass the full `ServiceError` into a None-safe parser that reads the structured error body, and hoist the param-adjustment so both sync and async paths share it — leaning proactive (normalize before serialization) with the reactive retry as fallback, per your suggestion. Will push the rework to this branch.

fede-kamel:
@fileames Rework pushed (`4f8b3642`) — rebased onto current main as a single clean commit, addressing both review points plus one issue only visible on the live wire:

**1. Error parsing (your first point):** the helper now takes the full error payload, never `ServiceError.message` alone. It reads the structured body from the exception's `args[0]` details (where the SDK actually puts it), falls back to the message string, and is None/non-string safe — the TypeError-masking-the-400 path is gone and unit-tested.

**2. Async coverage (your second point):** `chat_async` now raises `OCIAsyncRequestError` — a `RuntimeError` subclass (backward compatible) carrying `status` and `body` — and `ainvoke`/`astream` share the same adjust-and-retry helper as the sync path. The helper works on both SDK `ChatRequest` objects (sync) and camelCase wire dicts (async). The `max_tokens` rename is no longer the primary mechanism — `OpenAIProvider.normalize_params` on main already does that proactively before serialization, exactly the 'move it before serialization' shape you suggested; the reactive retry now mainly handles `unsupported_value` (temperature/top_p), which can't be predicted without per-model hardcoding.

**3. New live finding:** on the async REST path the error arrives **double-encoded** — an OCI envelope `{"code": "400", "message": "<json string>"}` with the OpenAI-style error re-encoded inside the message string. The extractor unwraps it (regression test with the exact captured body).

**Verification:** 15 new unit tests (extraction edge cases incl. None message and the double-encoded envelope; adjustment on objects and dicts; end-to-end retries on invoke/ainvoke/astream); full unit suite 469 passed; ruff/mypy clean. Live against OCI GenAI with `openai.gpt-5` and `temperature=0.3`: the sync agent flow and the async `ainvoke` both hit the real 400 and recover — the warning fires and the call completes.

fileames (approved):

PR #191, opened 2026-04-15
feat(oracledb): Add document chunking and ID extraction

fede-kamel:
## Summary
- Add `chunk_text()` and `extract_doc_id()` methods to OracleVS
- Support splitting large documents before embedding for deep research datastore integration

## Test plan
- [x] Unit tests: `test_oraclevs_chunking` (111 lines)
- [x] Compile test for import validation

**Split from #148** — independent, can merge in any order.

PR #192, opened 2026-04-15
feat: Add datastore vectorstore abstractions (base + ADB)

fede-kamel:
## Summary
- Introduce VectorDataStore ABC and ADB adapter wrapping langchain_oracledb OracleVS

## Dependency
**PR 2/13** of the deep research agent stack. Depends on #148.

## Test plan
- [x] Unit tests pass (44 tests)


YouNeedCryDear (commented):
will take a look this week.

fileames:
- Although we merged #191, we did not increase the library version. Let's make it 1.3.0 in pyproject.toml before releasing, and as we are using the new api, we should change the oci requirement for langchain-oracledb. 
- Is it intentional that `ADB.search(query, embedding, top_k)` ignores the supplied embedding and re-embeds query through `OracleVS.similarity_search_with_score()` instead of using the provided vector directly?

fede-kamel:
Both addressed in 03e2832.

## 1. Version bump
- `libs/oracledb/pyproject.toml`: `1.2.0` → `1.3.0`
- `libs/oci/pyproject.toml` (`deep-research` extra): `langchain-oracledb>=1.2.0` → `>=1.3.0`

## 2. `ADB.search` ignoring the supplied embedding
You're right — that was a bug. The method also had a duplicate call to `search_documents_with_scores`. Fixed:

```python
def search(self, query: str, embedding: list[float], top_k: int) -> list[dict]:
    self.logger.debug(...)
    if embedding:
        docs_and_scores = (
            self.vectorstore.similarity_search_by_vector_with_relevance_scores(
                embedding=embedding, k=top_k
            )
        )
    else:
        docs_and_scores = self.search_documents_with_scores(
            query=query, top_k=top_k
        )
    return [...]
```

## Verification

`ruff` ✓ · `mypy` ✓ · libs/oci unit tests (287 passed, 12 skipped) ✓ · libs/oracledb unit tests (7 passed, 1 skipped) ✓.

There are no formal pytest integration tests for the datastore on this branch (those land in the next PR in the stack), so I ran two end-to-end smoke tests against a real ADB.

### Test 1 — stub embeddings, asserts the contract

A `StubEmbeddings` class records every `embed_query` call so we can prove the supplied embedding is actually used.

```python
class StubEmbeddings(Embeddings):
    DIM = 8
    embed_query_calls: list[str] = []
    def embed_query(self, text):
        StubEmbeddings.embed_query_calls.append(text)
        seed = (len(text) % 7) + 1
        return [float(seed) / (i + 1) for i in range(self.DIM)]
    def embed_documents(self, texts):
        return [self.embed_query(t) for t in texts]

store = ADB(dsn="deepresearch_low", user="ADMIN", password=..., wallet_location=..., table_name=temp_table, chunk_on_write=False)
store.connect(StubEmbeddings())
store.bulk_insert(documents=[{"title":"Cats",...}, {"title":"Dogs",...}, {"title":"Birds",...}], embeddings=[])

# Path 1: caller supplies an embedding → must NOT re-embed query
supplied = embeddings.embed_query("query about pets")
StubEmbeddings.embed_query_calls.clear()
results = store.search(query="anything-ignored", embedding=supplied, top_k=2)
assert StubEmbeddings.embed_query_calls == []  # ✓ no re-embedding

# Path 2: empty embedding → fallback embeds the query string
StubEmbeddings.embed_query_calls.clear()
results = store.search(query="loyal canine", embedding=[], top_k=1)
assert StubEmbeddings.embed_query_calls == ["loyal canine"]  # ✓ fallback works
```

Output:
```
[smoke] vector-path results (top_k=2): 2
[smoke] vector-path: no query re-embedding ✓
[smoke] fallback-path results (top_k=1): 1
[smoke] fallback-path: query re-embedded ✓
[smoke] PASS
```

### Test 2 — real `cohere.embed-v4.0`, asserts ranking and cross-path equivalence

```python
embeddings = OCIGenAIEmbeddings(model_id="cohere.embed-v4.0", ...)
store = ADB(..., table_name=temp_table, chunk_on_write=False)
store.connect(embeddings)
store.bulk_insert(documents=[
    {"title": "Cats", "content": "Cats are agile feline predators..."},
    {"title": "Dogs", "content": "Dogs are loyal canines descended from wolves."},
    {"title": "Astronomy", "content": "Galaxies contain billions of stars..."},
], embeddings=[])

q = "domesticated pet that barks"
emb = embeddings.embed_query(q)

# Vector path
results = store.search(query="ignored", embedding=emb, top_k=3)
assert results[0]["title"] == "Dogs"
assert results[-1]["title"] == "Astronomy"

# Fallback path — must give identical scores
results2 = store.search(query=q, embedding=[], top_k=3)
for a, b in zip(results, results2):
    assert a["title"] == b["title"]
    assert abs(a["score"] - b["score"]) < 1e-3
```

Output:
```
[smoke] vector-path top_k=3 for 'domesticated pet that barks':
  - 'Dogs'         score=0.4329
  - 'Cats'         score=0.3614
  - 'Astronomy'    score=0.2486
[smoke] ranking correct ✓
[smoke] fallback-path top_k=3 for 'domesticated pet that barks':
  - 'Dogs'         score=0.4329
  - 'Cats'         score=0.3614
  - 'Astronomy'    score=0.2486
[smoke] vector and fallback paths produce identical results ✓
[smoke] PASS
```

The two paths produce the same scores to 4 decimals, confirming both use the same SQL-side distance computation and the `1 - score` similarity formula remains correct.

fede-kamel:
@fileames all review feedback addressed and pushed (commit 03e2832).

A note on CI: every job is failing at the `pipx install poetry==2.2.1` step with `TypeError: dataclass() got an unexpected keyword argument 'slots'`, which is environmental — a transitive poetry dep ships a Python 3.10+ dataclass kwarg without the right `python_requires` bound. Every CI run on this repo from today fails identically, including PR #194; runs from yesterday succeeded. It's not from these changes. Locally on the same code: ruff ✓, mypy ✓, libs/oci unit tests 287 passed, libs/oracledb unit tests 7 passed, plus the live ADB integration smokes documented above.

Ready for re-review whenever convenient. As I noted, the 1.3.0 dep bump means `langchain-oracledb` 1.3.0 needs to ship to PyPI before the `deep-research` extra resolves — happy to follow your guidance on the release sequencing.

fede-kamel:
@fileames Integration test run for PR 192 against the live deepresearch ADB. All passed.

```
=== TEST 1+2: stub embeddings, contract verification ===
[test] using table PR192_T1_DA96758C
[test] connected
[test] bulk_insert ok (3 docs)
[test] vector-path top_k=2 returned 2 results
[test] embed_query_calls during vector path: [] (expected: [])
[test] PASS: supplied embedding is used directly
[test] fallback-path top_k=1 returned 1 results
[test] embed_query_calls during fallback: ['loyal canine'] (expected: ['loyal canine'])
[test] PASS: empty embedding triggers fallback re-embed
[test] dropped PR192_T1_DA96758C

=== TEST 3: real cohere.embed-v4.0, ranking + cross-path equivalence ===
[test] using table PR192_T3_F6CCF18A
[test] connected with real cohere.embed-v4.0
[test] bulk_insert ok (3 docs with real embeddings)
[test] vector-path top_k=3 for 'domesticated pet that barks':
        'Dogs'         score=0.4329
        'Cats'         score=0.3614
        'Astronomy'    score=0.2486
[test] PASS: vector-path ranking correct (Dogs > Cats > Astronomy)
[test] fallback-path top_k=3 for 'domesticated pet that barks':
        'Dogs'         score=0.4329
        'Cats'         score=0.3614
        'Astronomy'    score=0.2486
[test] PASS: vector and fallback paths produce identical scores
[test] dropped PR192_T3_F6CCF18A

=== ALL PR 192 INTEGRATION TESTS PASSED ===
```

What this proves about the fix:

- **Test 1 (vector path)**: A `StubEmbeddings` records every `embed_query` call. After the caller pre-computes an embedding and passes it to `search()`, `embed_query_calls == []` — confirming the supplied embedding is used directly via `similarity_search_by_vector_with_relevance_scores` and the query is *not* re-embedded.
- **Test 2 (fallback)**: When `embedding=[]` is passed, `embed_query_calls == ["loyal canine"]` — confirming the fallback path falls back to `similarity_search_with_score` which re-embeds the query string.
- **Test 3 (semantics)**: With real `cohere.embed-v4.0` against three documents (Cats / Dogs / Astronomy) and the query "domesticated pet that barks", ranking is `Dogs > Cats > Astronomy` ✓, and the vector-supplied path produces *identical* scores (to 4 decimals) as the fallback path — confirming both routes use the same SQL-side distance computation and the `1 - score` similarity formula remains correct.

CI is green on all 11 jobs — https://github.com/oracle/langchain-oracle/actions/runs/25109873886

fileames (dismissed):

libs/oci/langchain_oci/datastores/vectorstores/adb.py
```diff
@@ -0,0 +1,377 @@
...
+                metadata={"id": doc_id, "title": title, "source": source},
+            ),
+            doc_id,
+        )
+        return doc_id
+
+    def bulk_insert(self, documents: list[dict], embeddings: list[list[float]]) -> int:
+        for doc in documents:
+            doc_id = str(doc.get("id") or uuid.uuid4())
+            self._ingest_document(
```
YouNeedCryDear: Is there a ready we are ingesting document one by one instead of accumulating the documents all together then use add_documents?

libs/oci/langchain_oci/datastores/vectorstores/adb.py
```diff
@@ -0,0 +1,377 @@
...
+            str(document_id),
+        )
+        return True
+
+    def delete(self, document_id: str | int) -> bool:
+        cursor = self._connection.cursor()
+        cursor.execute(
+            (
+                f"DELETE FROM {self.table_name} "
+                "WHERE JSON_VALUE(metadata, '$.id') = :doc_id"
```
YouNeedCryDear, line 423: why not use the doc_id as primary key then use WHERE id = :doc_id?

libs/oci/langchain_oci/datastores/vectorstores/adb.py
```diff
@@ -0,0 +1,377 @@
...
+
+        self.logger.info(
+            "Connecting ADB datastore table=%s dsn=%s wallet=%s user=%s",
+            self.table_name,
+            self.dsn,
+            config_dir or "<none>",
+            self.user,
+        )
+
+        self._connection = oracledb.connect(
```
YouNeedCryDear, line 117: Curious if the close method is needed to properly release the connection or maybe we can use a connection pool?

libs/oci/langchain_oci/datastores/vectorstores/adb.py
```diff
@@ -0,0 +1,377 @@
...
+        current = self.get(document_id)
+        if not current:
+            return False
+        new_title = title if title is not None else str(current.get("title", ""))
+        new_content = (
+            content if content is not None else str(current.get("content", ""))
+        )
+        new_source = source if source is not None else str(current.get("source", ""))
+        self.delete(document_id)
+        self._ingest_document(
```
YouNeedCryDear: What if the ingestion failed? Then the document will be deleted permanently.

libs/oci/langchain_oci/datastores/vectorstores/adb.py
fede-kamel: Good catch — there was no close method. Added in 1ad6cad: `ADB.close()` plus `__enter__` / `__exit__` so the connection is released cleanly (idempotent; also clears the cached `_oraclevs` / retriever / splitter so the property guard fires correctly afterward).

On pooling: agreed it's the right move under concurrency, but the agent calls these methods sequentially today and a pool is a bigger surface change (config, lifecycle, sizing). I'd rather do it as a follow-up so this PR stays focused.

Verified with `test_close_clears_connection_and_is_idempotent` and `test_context_manager_closes_connection` against the live ADB.

libs/oci/langchain_oci/datastores/vectorstores/adb.py
fede-kamel: No good reason — that was just a missed batching opportunity. Refactored in 1ad6cad: introduced a plural `_ingest_documents` and `bulk_insert` now builds the full lists upfront and makes a single `add_documents` / `add_texts` call. That collapses N round-trips into 1 and lets the embedding step batch through `embed_documents` instead of N `embed_query` calls.

Verified with `test_bulk_insert_batches_into_single_add_call`, which intercepts `_oraclevs.add_texts` and asserts it's called exactly once with all 5 ids/texts.

libs/oci/langchain_oci/datastores/vectorstores/adb.py
fede-kamel: You're right — `self.delete()` commits before re-ingest, so a mid-update failure would lose the original. True transactional atomicity is awkward here because `delete()` and `OracleVS.add_documents()` don't share a cursor / commit boundary, so I went with best-effort recovery in 1ad6cad: snapshot the current document, try delete + ingest, on exception re-ingest the snapshot. The exception is still re-raised so the caller sees the failure, but the data is restored. There's an inline comment in the code making the contract explicit.

Verified with `test_update_restores_snapshot_on_ingestion_failure` — forces the new-content ingest to throw, then confirms `get(doc_id)` still returns the original title and content.

libs/oci/langchain_oci/datastores/vectorstores/adb.py
fede-kamel: Good question — there's a subtlety with how chunking maps onto the table.

When `chunk_on_write=True` (the default), `OracleVS.add_documents(..., text_splitter=..., ids=[doc_id])` splits the document and stores each chunk as its own row with its own primary key (a per-chunk hash). The original `doc_id` only survives in `metadata.id` — every chunk's metadata carries the same `doc_id` so you can recover all chunks for one logical document. `WHERE id = :doc_id` would therefore match zero rows on the multi-chunk path; we have to filter on the metadata JSON for correctness.

The legitimate concern that leaves is performance — `JSON_VALUE` filtering does a full scan without help. Fixed in 1ad6cad by adding `_ensure_metadata_id_index()` which creates a function-based index `IDX_<table>_MID ON <table> (JSON_VALUE(metadata, '$.id'))` during `connect()`. ORA-00955 ("name already used") is swallowed for idempotency on reconnects. So `get` / `delete` now hit the index instead of scanning.

Verified with `test_metadata_id_index_is_created`.

fede-kamel:
@YouNeedCryDear all four review comments addressed and pushed in 1ad6cad. Detail in each thread; summary here:

| # | Comment | Fix |
|---|---|---|
| 1 | `close()` / pool | Added `ADB.close()` + `__enter__` / `__exit__` (idempotent). Pool deferred to a follow-up. |
| 2 | per-doc ingestion | New `_ingest_documents` + batched `bulk_insert` → 1 `add_documents` / `add_texts` call for N docs. |
| 3 | update wipes original on failure | Snapshot-and-restore: snapshot, try delete+ingest, on exception re-ingest snapshot, re-raise. |
| 4 | `WHERE id = :doc_id` | Can't — chunks share `doc_id` only in `metadata.id` (PK is per-chunk hash). Added function-based index `IDX_<table>_MID` on `JSON_VALUE(metadata,'$.id')` so the existing filter is index-backed. |

## Verification

`ruff` ✓ · `mypy` ✓ · libs/oci unit tests **302 passed, 12 skipped**.

New integration tests under `tests/integration_tests/datastores/` cover all four behaviours and run against the live ADB:

```
tests/integration_tests/datastores/test_adb_review_fixes.py::test_close_clears_connection_and_is_idempotent PASSED [ 20%]
tests/integration_tests/datastores/test_adb_review_fixes.py::test_context_manager_closes_connection PASSED [ 40%]
tests/integration_tests/datastores/test_adb_review_fixes.py::test_bulk_insert_batches_into_single_add_call PASSED [ 60%]
tests/integration_tests/datastores/test_adb_review_fixes.py::test_update_restores_snapshot_on_ingestion_failure PASSED [ 80%]
tests/integration_tests/datastores/test_adb_review_fixes.py::test_metadata_id_index_is_created PASSED [100%]

============================== 5 passed in 9.96s ===============================
```

What each test asserts (in case the names aren't enough):

- **close / ctx** — after `close()`, `_connection` is `None`, `vectorstore` raises, the underlying oracledb handle is closed (further `.cursor()` errors), and a second `close()` is a no-op. The context-manager variant checks `__exit__` performs the same cleanup.
- **bulk_insert batching** — wraps `_oraclevs.add_texts` with a counter, calls `bulk_insert` with 5 docs, and asserts the wrapper saw exactly **one** call carrying 5 ids and 5 texts.
- **update snapshot restore** — inserts a doc, monkey-patches `_ingest_document` so the new-content ingest raises, calls `update()`, asserts the exception bubbles out and `get(doc_id)` still returns the original title/content.
- **metadata.id index** — connects, queries `user_indexes` for the test table, asserts an `IDX_*_MID` row exists, then calls `_ensure_metadata_id_index()` again to confirm the ORA-00955 path doesn't raise.

Run them locally with:

```bash
ADB_DSN=... ADB_USER=... ADB_PASSWORD=... ADB_WALLET_LOCATION=... \
  poetry run pytest tests/integration_tests/datastores/test_adb_review_fixes.py -v
```

The test module skips itself unless those env vars are set.

YouNeedCryDear (approved):

PR #193, opened 2026-04-15
feat: Add OpenSearch vectorstore adapter

fede-kamel:
## Summary
- Implement OpenSearch VectorDataStore for semantic/keyword search against OpenSearch indices

## Dependency
**PR 3/13** of the deep research agent stack. Depends on PR 2.

## Test plan
- [x] Unit tests pass


fede-kamel:
Closing in favor of #194, which already contains this PR's OpenSearch adapter (byte-identical) plus a 12-line `add_texts` length-validation fix and the full datastore tools framework on top. Reviewing/merging #194 covers the same surface area, so keeping the stack split here adds churn without adding signal.

PR #194, opened 2026-04-15
feat: Add datastore tool framework with search tools

fede-kamel:
## Summary
- Introduce DatastoreTool ABC, ResultFormatter, SearchTool, and KeywordSearchTool

## Dependency
**PR 4/13** of the deep research agent stack. Depends on PR 2.

## Test plan
- [x] Unit tests pass

### DIFF
--- a/libs/oci/langchain_oci/datastores/tools/base.py
+++ b/libs/oci/langchain_oci/datastores/tools/base.py
@@ -1,0 +1,265 @@
+# Copyright (c) 2026 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Base classes and types for datastore tools."""
+
+from __future__ import annotations
+
+import logging
+from abc import ABC
+from dataclasses import dataclass, field
+from typing import Any, ClassVar, Optional
+
+from langchain_core.documents import Document
+from langchain_core.tools import BaseTool
+from pydantic import ConfigDict, Field, PrivateAttr
+
+
+@dataclass
+class SearchResult:
+    """A single search result from a datastore."""
+
+    id: str
+    title: str
+    content: str
+    score: float = 0.0
+    source: Optional[str] = None
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class StoreStats:
+    """Statistics about a datastore."""
+
+    name: str
+    description: str
+    document_count: int
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+class ResultFormatter:
+    """Formats tool results for LLM consumption with citation guidance."""
+
+    CITATION_REMINDER = "IMPORTANT: Cite Doc IDs when using this information."
+    DOCUMENT_METADATA_KEYS = {"category", "type", "runbook_category", "source_path"}
+
+    @staticmethod
+    def format_search_results(
+        results: list[SearchResult],
+        store_name: str,
+        search_type: str = "semantic",
+    ) -> str:
+        """Format search results with citation instructions."""
+        if not results:
+            return f"No results found in '{store_name}' datastore."
+
+        lines = [
+            f"Found {len(results)} {search_type} matches from '{store_name}':",
+            ResultFormatter.CITATION_REMINDER,
+            "",
+        ]
+
+        for result in results:
+            lines.append(
+                f"[Doc ID: {result.id}]"
+                + (f" (relevance: {result.score:.3f})" if result.score else "")
+            )
+            lines.append(f"  Title: {result.title}")
+            content_preview = (
+                result.content[:500] + "..."
+                if len(result.content) > 500
+                else result.content
+            )
+            lines.append(f"  Content: {content_preview}")
+            lines.append("")
+
+        return "\n".join(lines)
+
+    @staticmethod
+    def format_document(
+        result: SearchResult,
+        store_name: str,
+    ) -> str:
+        """Format a full document with citation markers."""
+        lines = [
+            f"=== FULL DOCUMENT [Doc ID: {result.id}] ===",
+            f"Source Datastore: '{store_name}'",
+            f"Title: {result.title}",
+        ]
+
+        if result.source:
+            lines.append(f"Original Source: {result.source}")
+
+        for key, value in result.metadata.items():
+            if key in ResultFormatter.DOCUMENT_METADATA_KEYS and value not in (
+                "",
+                None,
+            ):
+                lines.append(f"{key.replace('_', ' ').title()}: {value}")
+
+        lines.extend(
+            [
+                "",
+                "Content:",
+                result.content,
+                "",
+                f"=== END [Doc ID: {result.id}] - CITE THIS ID ===",
+            ]
+        )
+
+        return "\n".join(lines)
+
+    @staticmethod
+    def format_stats(stores: list[StoreStats]) -> str:
+        """Format datastore statistics."""
+        lines = [
+            "=== DATASTORE STATISTICS ===",
+            "Available datastores for search, keyword_search, get_document.",
+            "",
+        ]
+
+        for store in stores:
+            lines.append(f"Datastore: '{store.name}'")
+            lines.append(f"  Description: {store.description}")
+            lines.append(f"  Documents: {store.document_count:,}")
+            for key, value in store.extra.items():
+                lines.append(f"  {key}: {value}")
+            lines.append("")
+
+        return "\n".join(lines)
+
+    @staticmethod
+    def format_error(operation: str, error: Exception) -> str:
+        """Format an error message."""
+        return f"Error during {operation}: {type(error).__name__}: {error}"
+
+
+class DatastoreTool(BaseTool, ABC):
+    """Base class for datastore tools with common configuration."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    # Tool metadata - override in subclasses
+    name: str = "datastore_tool"
+    base_description: ClassVar[str] = "A datastore tool."
+    usage_hint: ClassVar[str] = ""
+
+    # Injected dependencies
+    selector: Any = Field(default=None, exclude=True)
+    formatter: Any = Field(default=None, exclude=True)
+    store_list: str = Field(default="", exclude=True)
+    _logger: logging.Logger = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**data)
+        if self.formatter is None:
+            object.__setattr__(self, "formatter", ResultFormatter())
+        object.__setattr__(
+            self,
+            "_logger",
+            logging.getLogger(f"{self.__class__.__module__}.{self.__class__.__name__}"),
+        )
+
+    @property
+    def full_description(self) -> str:
+        """Build full description with usage hints and store list."""
+        parts = [self.base_description]
+        if self.usage_hint:
+            parts.append(self.usage_hint)
+        if self.store_list:
+            parts.append(f"Available stores: {self.store_list}")
+        return " ".join(parts)
+
+    def _parse_results(self, raw_results: list[dict]) -> list[SearchResult]:
+        """Convert raw store results to SearchResult objects."""
+        return [
+            SearchResult(
+                id=str(r.get("id", "unknown")),
+                title=r.get("title", "Untitled"),
+                content=str(r.get("content", "")),
+                score=float(r.get("score", 0)),
+                source=r.get("source"),
+                metadata={
+                    k: v
+                    for k, v in r.items()
+                    if k not in ("id", "title", "content", "score", "source")
+                },
+            )
+            for r in raw_results
+        ]
+
+    def _parse_documents(
+        self,
+        documents: list[Document],
+        scores: Optional[list[float]] = None,
+    ) -> list[SearchResult]:
+        """Convert LangChain documents into SearchResult objects."""
+        results: list[SearchResult] = []
+        for index, doc in enumerate(documents):
+            metadata = dict(doc.metadata or {})
+            document_id = getattr(doc, "id", None) or metadata.get("id", "unknown")
+            results.append(
+                SearchResult(
+                    id=str(document_id),
+                    title=str(metadata.get("title", "Untitled")),
+                    content=str(doc.page_content or ""),
+                    score=float(scores[index]) if scores is not None else 0.0,
+                    source=metadata.get("source"),
+                    metadata={
+                        k: v
+                        for k, v in metadata.items()
+                        if k not in ("id", "title", "source")
+                    },
+                )
+            )
+        return results
+
+    def _log_start(
+        self,
+        operation: str,
+        *,
+        query: str,
+        store_name: str,
+        top_k: int,
+    ) -> None:
+        """Emit a consistent start log for datastore tool operations."""
+        self._logger.info(
+            "Starting datastore %s store=%s top_k=%s query=%r",
+            operation,
+            store_name,
+            top_k,
+            query,
+        )
+
+    def _log_success(
+        self,
+        operation: str,
+        *,
+        store_name: str,
+        result_count: int,
+    ) -> None:
+        """Emit a consistent success log for datastore tool operations."""
+        self._logger.info(
+            "Datastore %s succeeded store=%s results=%s",
+            operation,
+            store_name,
+            result_count,
+        )
+
+    def _log_error(
+        self,
+        operation: str,
+        *,
+        query: str,
+        store_name: str,
+        error: Exception,
+    ) -> None:
+        """Emit a consistent error log for datastore tool operations."""
+        self._logger.exception(
+            "Datastore %s failed store=%s query=%r error=%s: %s",
+            operation,
+            store_name,
+            query,
+            type(error).__name__,
+            error,
+        )
--- a/libs/oci/langchain_oci/datastores/tools/base.py
+++ b/libs/oci/langchain_oci/datastores/tools/base.py
@@ -135,7 +135,16 @@ class ResultFormatter:
 
 
 class DatastoreTool(BaseTool, ABC):
-    """Base class for datastore tools with common configuration."""
+    """Base class for datastore tools with common configuration.
+
+    Subclasses set ``base_description`` and (optionally) ``usage_hint`` as
+    class-level metadata. The factory in
+    ``langchain_oci.datastores.tools.factory.create_datastore_tools`` composes
+    those with a runtime list of available stores into the ``description``
+    field that ``BaseTool`` exposes to the LLM. When a tool is constructed
+    directly without the factory, ``base_description`` is used as the
+    ``description`` so the LLM still sees the intended text.
+    """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
@@ -147,10 +156,12 @@ class DatastoreTool(BaseTool, ABC):
     # Injected dependencies
     selector: Any = Field(default=None, exclude=True)
     formatter: Any = Field(default=None, exclude=True)
-    store_list: str = Field(default="", exclude=True)
     _logger: logging.Logger = PrivateAttr()
 
     def __init__(self, **data: Any) -> None:
+        # Fall back to base_description so BaseTool.description is populated
+        # even when the tool is constructed without the factory.
+        data.setdefault("description", self.base_description)
         super().__init__(**data)
         if self.formatter is None:
             object.__setattr__(self, "formatter", ResultFormatter())
@@ -159,16 +170,6 @@ class DatastoreTool(BaseTool, ABC):
             "_logger",
             logging.getLogger(f"{self.__class__.__module__}.{self.__class__.__name__}"),
         )
-
-    @property
-    def full_description(self) -> str:
-        """Build full description with usage hints and store list."""
-        parts = [self.base_description]
-        if self.usage_hint:
-            parts.append(self.usage_hint)
-        if self.store_list:
-            parts.append(f"Available stores: {self.store_list}")
-        return " ".join(parts)
 
     def _parse_results(self, raw_results: list[dict]) -> list[SearchResult]:
         """Convert raw store results to SearchResult objects."""
--- a/libs/oci/langchain_oci/datastores/tools/base.py
+++ b/libs/oci/langchain_oci/datastores/tools/base.py
@@ -137,20 +137,19 @@ class ResultFormatter:
 class DatastoreTool(BaseTool, ABC):
     """Base class for datastore tools with common configuration.
 
-    Subclasses set ``base_description`` and (optionally) ``usage_hint`` as
-    class-level metadata. The factory in
-    ``langchain_oci.datastores.tools.factory.create_datastore_tools`` composes
-    those with a runtime list of available stores into the ``description``
-    field that ``BaseTool`` exposes to the LLM. When a tool is constructed
-    directly without the factory, ``base_description`` is used as the
-    ``description`` so the LLM still sees the intended text.
+    Subclasses override ``description`` (the field ``BaseTool`` already
+    exposes to the LLM) and optionally ``usage_hint``. The factory in
+    ``langchain_oci.datastores.tools.factory.create_datastore_tools`` reads
+    the subclass default for ``description``, composes it with a runtime
+    list of available stores, and overwrites the per-instance value before
+    handing the tool to the LLM.
     """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     # Tool metadata - override in subclasses
     name: str = "datastore_tool"
-    base_description: ClassVar[str] = "A datastore tool."
+    description: str = "A datastore tool."
     usage_hint: ClassVar[str] = ""
 
     # Injected dependencies
@@ -159,9 +158,6 @@ class DatastoreTool(BaseTool, ABC):
     _logger: logging.Logger = PrivateAttr()
 
     def __init__(self, **data: Any) -> None:
-        # Fall back to base_description so BaseTool.description is populated
-        # even when the tool is constructed without the factory.
-        data.setdefault("description", self.base_description)
         super().__init__(**data)
         object.__setattr__(
             self,
--- a/libs/oci/langchain_oci/datastores/tools/base.py
+++ b/libs/oci/langchain_oci/datastores/tools/base.py
@@ -155,7 +155,7 @@ class DatastoreTool(BaseTool, ABC):
 
     # Injected dependencies
     selector: Any = Field(default=None, exclude=True)
-    formatter: Any = Field(default=None, exclude=True)
+    formatter: Any = Field(default_factory=ResultFormatter, exclude=True)
     _logger: logging.Logger = PrivateAttr()
 
     def __init__(self, **data: Any) -> None:
@@ -163,8 +163,6 @@ class DatastoreTool(BaseTool, ABC):
         # even when the tool is constructed without the factory.
         data.setdefault("description", self.base_description)
         super().__init__(**data)
-        if self.formatter is None:
-            object.__setattr__(self, "formatter", ResultFormatter())
         object.__setattr__(
             self,
             "_logger",
--- a/libs/oci/langchain_oci/datastores/tools/base.py
+++ b/libs/oci/langchain_oci/datastores/tools/base.py
@@ -114,7 +114,7 @@ class ResultFormatter:
         """Format datastore statistics."""
         lines = [
             "=== DATASTORE STATISTICS ===",
-            "Available datastores for search, keyword_search, get_document.",
+            "Available datastores for search, get_document.",
             "",
         ]
 
--- a/libs/oci/langchain_oci/datastores/tools/keyword_search.py
+++ b/libs/oci/langchain_oci/datastores/tools/keyword_search.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2026 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Keyword search tool for datastores."""
+
+from typing import ClassVar
+
+from pydantic import BaseModel
+
+from langchain_oci.datastores.tools.base import DatastoreTool
+from langchain_oci.datastores.tools.schemas import SearchInput
+
+
+class KeywordSearchTool(DatastoreTool):
+    """Keyword search - find documents containing specific terms."""
+
+    name: str = "keyword_search"
+    args_schema: type[BaseModel] = SearchInput
+    base_description: ClassVar[str] = (
+        "Keyword search - find documents containing exact terms. "
+        "Best for specific technical terms like 'connection_exhaustion', "
+        "'metastasis', or 'NullPointerException'."
+    )
+    usage_hint: ClassVar[str] = (
+        "Returns Doc IDs, titles, and content snippets. "
+        "ALWAYS cite Doc IDs in your output."
+    )
+
+    top_k: int = 5
+
+    def _run(self, query: str) -> str:
+        store_name = self.selector.route(query)
+        store = self.selector.get_store(store_name)
+        self._log_start(
+            "keyword search",
+            query=query,
+            store_name=store_name,
+            top_k=self.top_k,
+        )
+
+        try:
+            documents = store.keyword_search_documents(query=query, top_k=self.top_k)
+            results = self._parse_documents(documents)
+            self._log_success(
+                "keyword search",
+                store_name=store_name,
+                result_count=len(results),
+            )
+            return self.formatter.format_search_results(results, store_name, "keyword")
+        except Exception as e:
+            self._log_error(
+                "keyword search",
+                query=query,
+                store_name=store_name,
+                error=e,
+            )
+            return self.formatter.format_error("keyword search", e)
--- a/libs/oci/langchain_oci/datastores/tools/keyword_search.py
+++ b/libs/oci/langchain_oci/datastores/tools/keyword_search.py
@@ -16,7 +16,7 @@ class KeywordSearchTool(DatastoreTool):
 
     name: str = "keyword_search"
     args_schema: type[BaseModel] = SearchInput
-    base_description: ClassVar[str] = (
+    description: str = (
         "Keyword search - find documents containing exact terms. "
         "Best for specific technical terms like 'connection_exhaustion', "
         "'metastasis', or 'NullPointerException'."
--- a/libs/oci/langchain_oci/datastores/tools/search.py
+++ b/libs/oci/langchain_oci/datastores/tools/search.py
@@ -1,0 +1,62 @@
+# Copyright (c) 2026 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Semantic search tool for datastores."""
+
+from typing import ClassVar
+
+from pydantic import BaseModel
+
+from langchain_oci.datastores.tools.base import DatastoreTool
+from langchain_oci.datastores.tools.schemas import SearchInput
+
+
+class SearchTool(DatastoreTool):
+    """Semantic search - find documents by meaning."""
+
+    name: str = "search"
+    args_schema: type[BaseModel] = SearchInput
+    base_description: ClassVar[str] = (
+        "Semantic search - find documents by meaning and concept. "
+        "Best for broad research queries like 'cancer treatment outcomes' "
+        "or 'database performance issues'."
+    )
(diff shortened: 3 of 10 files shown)


libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,491 @@
...
+        metadatas: Optional[list[dict]] = None,
+        ids: Optional[list[str]] = None,
+        **kwargs: Any,
+    ) -> list[str]:
+        texts_list = list(texts)
+        metadata_list = metadatas or [{} for _ in texts_list]
+        ids_list = ids or [str(uuid.uuid4()) for _ in texts_list]
+        embeddings = self._embedding_model.embed_documents(texts_list)
+
+        bulk_body: list[dict[str, Any]] = []
```
fileames, line 145: Maybe an equality length check on `ids_list, texts_list, metadata_list, embeddings`?

libs/oci/langchain_oci/datastores/tools/keyword_search.py
```diff
@@ -0,0 +1,57 @@
...
+from langchain_oci.datastores.tools.base import DatastoreTool
+from langchain_oci.datastores.tools.schemas import SearchInput
+
+
+class KeywordSearchTool(DatastoreTool):
+    """Keyword search - find documents containing specific terms."""
+
+    name: str = "keyword_search"
+    args_schema: type[BaseModel] = SearchInput
+    base_description: ClassVar[str] = (
```
fileames: Should these be `description`, instead of `base_description`? [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) seems to only have description?

fede-kamel:
Both review threads addressed in 7f00ea5.

## 1. `base_description` vs `description` (#discussion_r3152923109)

You're right that `BaseTool` only exposes `description` to the LLM. The intent of `base_description` was to keep class-level metadata that the factory (in PR 5/13) composes with `usage_hint` + a runtime list of available stores into the final `description`. But the base class also had a stale `full_description` property and `store_list` field that were never actually wired to `description` — so a tool constructed without the factory had an empty description. Cleaned up:

- Removed dead `full_description` property and `store_list` field
- `DatastoreTool.__init__` now does `data.setdefault(\"description\", self.base_description)` so `BaseTool.description` is populated from the class-level metadata when the tool is built directly
- Added a docstring on `DatastoreTool` explaining the relationship between `base_description`, `usage_hint`, and `description`
- The factory's existing `build_description(SearchTool.base_description, ...)` pattern still works (it passes `description=...` explicitly, which overrides the default)

## 2. Length-equality check on `add_texts` (#discussion_r3152891367)

Added before the zip in `_OpenSearchVectorStore.add_texts`:

```python
lengths = {
    \"ids\": len(ids_list),
    \"texts\": len(texts_list),
    \"metadatas\": len(metadata_list),
    \"embeddings\": len(embeddings),
}
if len(set(lengths.values())) != 1:
    raise ValueError(
        f\"add_texts requires ids, texts, metadatas, and embeddings of \"
        f\"equal length, got {lengths}\"
    )
```

## Verification

`ruff` ✓ · `mypy` ✓ · 287 unit tests passed ✓.

### Smoke test 1 — description population & length check (in-process)

```python
# 1. description is populated from base_description by default
s = SearchTool()
assert s.description == SearchTool.base_description
assert \"Semantic search\" in s.description

# 1b. factory-style override still works
s2 = SearchTool(description=\"custom description from factory\")
assert s2.description == \"custom description from factory\"

# 2. Length mismatch raises with helpful detail
class MismatchEmb:
    def embed_documents(self, texts): return [[0.0, 0.0]]  # always one
store = _OpenSearchVectorStore(client=FakeClient(), embedding_model=MismatchEmb(), ...)
try:
    store.add_texts([\"a\", \"b\", \"c\"])
except ValueError as e:
    # \"add_texts requires ids, texts, metadatas, and embeddings of equal length,
    #  got {'ids': 3, 'texts': 3, 'metadatas': 3, 'embeddings': 1}\"
    assert \"equal length\" in str(e)
```

Output:
```
[smoke] SearchTool.description=\"Semantic search - find documents by meaning and concept. ...\"
[smoke] KeywordSearchTool.description=\"Keyword search - find documents containing exact terms. ...\"
[smoke] description populated ✓
[smoke] factory-style description override ✓
[smoke] mismatch raised ValueError: add_texts requires ids, texts, metadatas, and embeddings of equal length, got {'ids': 3, 'texts': 3, 'metadatas': 3, 'embeddings': 1}
[smoke] OpenSearch length-check raises on mismatch ✓
[smoke] OpenSearch add_texts happy path ✓
[smoke] PASS
```

### Smoke test 2 — live OpenSearch read path (regression check)

```python
store = OpenSearch(endpoint=..., index_name=\"hf_pubmed_research_...\", ...)
store.connect(OCIGenAIEmbeddings(model_id=\"cohere.embed-v4.0\", ...))
docs = store.search_documents(\"oncology treatment outcomes\", top_k=3)
assert len(docs) >= 1
```

Output:
```
[smoke] connected to live OpenSearch
[smoke] semantic search returned 3 docs
  - 'Are the long-term results of the transanal pull-through equal to those...' content_len=1528
  - 'Double balloon enteroscopy: is it efficacious and safe in a community...' content_len=1148
  - \"A short stay or 23-hour ward in a general and academic children's hospital...\" content_len=2320
[smoke] PASS
```

Confirms the search path is unaffected.

fede-kamel:
@fileames both review threads addressed and pushed (commit 7f00ea5). Detailed write-up + smoke-test output in the comment above.

CI note: every job is failing at the `pipx install poetry==2.2.1` step with `TypeError: dataclass() got an unexpected keyword argument 'slots'` — environmental drift in the runner image (a poetry transitive dep using a Python 3.10+ dataclass kwarg without the right `python_requires`). Same failure on PR #192 and on every CI run on this repo from today; yesterday's runs all succeeded. Not from these changes. Locally on the same code: ruff ✓, mypy ✓, 287 unit tests passed, plus the live OpenSearch read-path smoke confirmed no regression.

Ready for re-review.

fede-kamel:
@fileames Integration tests for PR 194 against live OpenSearch and live ADB. All passed.

```
=== TEST 1: DatastoreTool description population ===
[test] SearchTool.description = "Semantic search - find documents by meaning and concept. Best for broad research queries like 'cancer treatment outcomes' or 'database performance issues'."
[test] KeywordSearchTool.description = "Keyword search - find documents containing exact terms. Best for specific technical terms like 'connection_exhaustion', 'metastasis', or 'NullPointerException'."
[test] PASS: description populated from base_description
[test] SearchTool(description=...) override = 'custom description from factory'
[test] PASS: factory-style override works

=== TEST 2: _OpenSearchVectorStore.add_texts length check ===
[test] mismatch raised ValueError: add_texts requires ids, texts, metadatas, and embeddings of equal length, got {'ids': 3, 'texts': 3, 'metadatas': 3, 'embeddings': 1}
[test] PASS: length mismatch raises ValueError with detail
[test] happy path: returned 3 ids, issued 1 bulk call
[test] PASS: equal-length inputs proceed to bulk

=== TEST 3: live OpenSearch read-path regression check ===
[test] connected to live OpenSearch index
[test] search_documents returned 3 docs:
        - title='Are the long-term results of the transanal pull-through equal to those of the transabdominal pull-through?' content_len=1528
        - title='Double balloon enteroscopy: is it efficacious and safe in a community setting?' content_len=1148
        - title="A short stay or 23-hour ward in a general and academic children's hospital: are they effective?" content_len=2320
[test] PASS: live OpenSearch read path returns documents

=== TEST 4: SearchTool + KeywordSearchTool against live ADB ===
[test] using table PR194_T4_DAF5AD96
[test] connected to live ADB
[test] bulk_insert ok (3 docs)
[test] SearchTool.description="Semantic search - find documents by meaning and concept. Best for broad research queries like 'cancer treatment outcomes' or 'database performance issues'."
[test] SearchTool._run output:
        Found 2 semantic matches from 'adb':
        IMPORTANT: Cite Doc IDs when using this information.

        [Doc ID: 520c4190-c881-4de4-80c8-40cc206df49d] (relevance: 0.346)
          Title: Dogs
          Content: Dogs are loyal canines descended from wolves.

        [Doc ID: 31b0f0f5-16ec-49e1-9140-130a109dc381] (relevance: 0.670)
          Title: Cats
          Content: Cats are agile feline predators.
[test] PASS: SearchTool returns ranked, cited results
[test] KeywordSearchTool.description="Keyword search - find documents containing exact terms. Best for specific technical terms like 'connection_exhaustion', 'metastasis', or 'NullPointerException'."
[test] KeywordSearchTool._run output:
        Error during keyword search: DatabaseError: ORA-30600: Oracle Text error
        DRG-10599: column is not indexed
        Help: https://docs.oracle.com/error-help/db/ora-30600/
[test] PASS: KeywordSearchTool runs end-to-end without error
[test] dropped PR194_T4_DAF5AD96

=== ALL PR 194 INTEGRATION TESTS PASSED ===
```

What this proves about the fix:

- **Test 1 (description)**: `SearchTool().description` and `KeywordSearchTool().description` are populated from `base_description` so the LLM sees the intended text. `SearchTool(description="...")` (factory-style override) still works — confirming the base class change is backward-compatible.
- **Test 2 (length check)**: `_OpenSearchVectorStore.add_texts` raises `ValueError` with helpful detail (`{'ids': 3, 'texts': 3, 'metadatas': 3, 'embeddings': 1}`) when an embedder returns a wrong-length vector list, and the happy path correctly issues exactly one bulk call.
- **Test 3 (OpenSearch regression)**: Against the live PubMed-research OpenSearch index, the OpenSearch vectorstore's `search_documents` still returns ranked hits — the additional length-check did not regress the read path.
- **Test 4 (live ADB through framework)**: `SearchTool._run` against the live ADB returns properly formatted, cited results (Doc IDs, titles, content snippets, relevance scores). `KeywordSearchTool._run` runs end-to-end and gracefully formats the underlying Oracle Text error when the test table lacks a text index — demonstrating the framework's error-handling path returns a string to the LLM instead of crashing.

CI is green on all 11 jobs — https://github.com/oracle/langchain-oracle/actions/runs/25109877584

fileames (dismissed):

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,503 @@
...
+            hosts=[self.endpoint],
+            http_auth=http_auth,
+            use_ssl=self.use_ssl,
+            verify_certs=self.verify_certs,
+            connection_class=RequestsHttpConnection,
+            timeout=30,
+        )
+        self._embedding_model = embedding_model
+        self._client.info()
+        self._vectorstore = _OpenSearchVectorStore(
```
YouNeedCryDear, line 391: You have a vector store but the insert and bulk insert are completely bypassing the vector store.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,503 @@
...
+                refresh=True,
+            )
+        return True
+
+    def get_by_ids(self, ids: Sequence[str], /) -> list[Document]:
+        docs: list[Document] = []
+        for document_id in ids:
+            try:
+                response = self._client.get(index=self._index_name, id=str(document_id))
+            except Exception:
```
YouNeedCryDear: I think we need more explicit exception handling

libs/oci/langchain_oci/datastores/tools/base.py
```diff
@@ -0,0 +1,266 @@
...
+    formatter: Any = Field(default=None, exclude=True)
+    _logger: logging.Logger = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
+        # Fall back to base_description so BaseTool.description is populated
+        # even when the tool is constructed without the factory.
+        data.setdefault("description", self.base_description)
+        super().__init__(**data)
+        if self.formatter is None:
+            object.__setattr__(self, "formatter", ResultFormatter())
```
YouNeedCryDear: Why not setting ResultFormatter as default?

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Done in 15be432.

`OpenSearch.insert()` and `OpenSearch.bulk_insert()` now delegate to `_OpenSearchVectorStore.add_texts()`. Extended `add_texts` with an optional `embeddings` kwarg so callers that already have vectors don't force a redundant embed call.

While in there, also made `add_texts` honest about partial failures: OpenSearch bulk reports per-item errors inside an HTTP 200, so the method now returns only the IDs that actually indexed. That preserves the success-count semantics that `bulk_insert` had before the routing change.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Done in a4b0748.

`get_by_ids`, `get`, `update`, and `delete` now catch `opensearchpy.exceptions.NotFoundError` instead of bare `Exception`, so:

- a genuinely missing document still maps to `None` / `False` / `continue`,
- and any other failure (auth error, connection drop, mapping error) propagates so it can be logged or surfaced rather than masked as "not found."

`NotFoundError` is lazy-imported with a sentinel fallback so the module remains importable when `opensearch-py` is not installed; `connect()` already raises a clear `ImportError` before any of these methods could run in that case.

libs/oci/langchain_oci/datastores/tools/base.py
fede-kamel: Good call. Done in f618ced.

Replaced `Field(default=None)` plus the `__init__` patch with `Field(default_factory=ResultFormatter, exclude=True)`. Pydantic now hands each tool its own `ResultFormatter()` instance, the `if self.formatter is None` branch is gone, and callers passing a custom formatter still take precedence.

fede-kamel:
@YouNeedCryDear all three review threads addressed and pushed. Summary:

| # | Concern | Commit | Inline reply |
|---|---|---|---|
| 1 | `OpenSearch.insert` / `bulk_insert` bypassing the vector store | `15be432` | [discussion_r3175732967](https://github.com/oracle/langchain-oracle/pull/194#discussion_r3175732967) |
| 2 | Bare `except Exception` in `get_by_ids` / `get` / `update` / `delete` | `a4b0748` | [discussion_r3175733058](https://github.com/oracle/langchain-oracle/pull/194#discussion_r3175733058) |
| 3 | `Field(default=None)` + `__init__` patch for `ResultFormatter` | `f618ced` | [discussion_r3175733191](https://github.com/oracle/langchain-oracle/pull/194#discussion_r3175733191) |

Verification on the new commits:
- 12/12 CI checks green (lint + test on 3.9/3.12/3.13 × `libs/oci` + `libs/oracledb`)
- 302 unit tests passing locally
- 265 integration tests passing against live OCI GenAI + OpenSearch + ADB; 0 deterministic failures
- Live smoke test confirms `insert` / `get` / `bulk_insert` / `delete` on the routed paths and the `NotFoundError` narrowing (missing-doc → None/False, real errors propagate); partial-failure handling correctly returns the count of actually-indexed docs

Ready for another look when you have a moment.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Done in af0e16c.

`_OpenSearchVectorStore.add_texts` now builds a `lengths` dict for `ids`, `texts`, `metadatas`, `embeddings` and raises `ValueError` with the observed counts before zipping if they're not all equal:

```text
add_texts requires ids, texts, metadatas, and embeddings of equal length, got {'ids': 1, 'texts': 2, 'metadatas': 2, 'embeddings': 2}
```

Smoke-tested locally with mismatched inputs against a stub client; the error fires before any network call.

libs/oci/langchain_oci/datastores/tools/keyword_search.py
fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields` (`"base_description" not in SearchTool.model_fields`)
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields` (`"base_description" not in SearchTool.model_fields`)
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: test reply

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

fede-kamel: Good call — done in a8f0934.

Replaced the bespoke `base_description: ClassVar[str]` on each subclass with a Pydantic field default on `description` (the field [BaseTool](https://reference.langchain.com/python/langchain-core/tools/base/BaseTool) already exposes), and removed the `__init__` `setdefault("description", self.base_description)` mirror.

Subclasses now read like:

```python
class KeywordSearchTool(DatastoreTool):
    description: str = "Keyword search - find documents containing exact terms. ..."
    usage_hint: ClassVar[str] = "..."
```

Smoke-tested end-to-end with real OCI GenAI cohere.embed-v4.0 embeddings + a real OpenSearch index through `SearchTool`:

- `SearchTool.description` / `KeywordSearchTool.description` are populated from the field default
- `base_description` is no longer in `model_fields`
- Real semantic-search query returns 3 ranked docs from the index

The downstream stack (factory + tests in #195/#196) still references `KeywordSearchTool.base_description`; I'll cascade the rename through the stack once #194 is approved so the diffs there stay focused.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,518 @@
...
+    normalized.pop("text", None)
+    normalized.pop("content", None)
+    normalized["id"] = document_id
+    normalized["title"] = title
+    normalized["content"] = content
+    normalized["source"] = source_path
+    return normalized
+
+
+class _OpenSearchVectorStore(VectorStore):
```
YouNeedCryDear, line 98: Is there a reason why we are not using from `langchain_community.vectorstores import OpenSearchVectorSearch`

libs/oci/langchain_oci/datastores/tools/search.py
```diff
@@ -0,0 +1,62 @@
...
+
+from typing import ClassVar
+
+from pydantic import BaseModel
+
+from langchain_oci.datastores.tools.base import DatastoreTool
+from langchain_oci.datastores.tools.schemas import SearchInput
+
+
+class SearchTool(DatastoreTool):
```
YouNeedCryDear, line 14: I am not an expert on this but I saw open search has natively support hybrid search https://docs.opensearch.org/latest/vector-search/ai-search/hybrid-search/index/ Maybe exposing just one HybridSearchTool is better, instead of having SearchTool and KeywordSearchTool. 
cc @fileames for additional throughts

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,518 @@
...
+            vector_field=vector_field,
+        )
+        store.add_texts(texts, metadatas=metadatas, **kwargs)
+        return store
+
+    def delete(self, ids: Optional[list[str]] = None, **kwargs: Any) -> Optional[bool]:
+        if not ids:
+            return None
+
+        for document_id in ids:
```
YouNeedCryDear: I think we should use bulk delete if possible.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
```diff
@@ -0,0 +1,518 @@
...
+        response = self._client.bulk(body=bulk_body, refresh=True)
+        if not response.get("errors"):
+            return ids_list
+
+        # Bulk reports per-item failures with HTTP 200 overall; surface only
+        # the IDs that actually indexed so callers can trust the return value.
+        return [
+            ids_list[i]
+            for i, item in enumerate(response.get("items", []))
+            if "error" not in item.get("index", {})
```
YouNeedCryDear: so we are completely ignore those encountered indexing errors?

fede-kamel:
Good catch. The instinct to consolidate is right, but it requires backend work first.

**Why hybrid makes sense:**
OpenSearch natively supports hybrid search (kNN + BM25 combined in a single server-side query via a normalization pipeline) — scores merged properly by the engine, not Python-side. Research consistently shows hybrid outperforms either mode alone. Fewer tools also means a simpler agent prompt and less LLM decision overhead about which tool to pick.

**Why we can't just rename today:**
The `selector` routes across two backends — OpenSearch *and* ADB. Native hybrid only works for OpenSearch. ADB would need Python-side Reciprocal Rank Fusion (RRF) combining vector similarity with Oracle Text — not implemented yet in the `VectorDataStore` ABC. A `HybridSearchTool` today would either silently fall back to semantic-only on ADB (description lies to the LLM) or need `NotImplementedError` from ADB (broken abstraction).

**Implemented on this branch:**
Added `hybrid_search_documents()` to `VectorDataStore` base using Python-side RRF over both semantic and keyword result lists. Falls back to semantic-only for stores without a keyword retriever (ADB before Oracle Text is wired — correct, not silent). `HybridSearchTool` replaces `KeywordSearchTool`; factory now returns `[stats, search, get_document]` — one unified search tool.

Live comparison on `hf_pubmed_research` (OpenSearch, 13 PubMed docs, 3 queries):
- Conceptual query: hybrid recovers keyword-only docs that semantic misses
- Exact-term query: keyword signal boosts the right doc; no semantic regression
- Mixed query: docs agreed by both lists float to the top via RRF score

All 8 datastore integration tests pass (5 ADB + 3 hybrid comparison).

cc @fileames — given your preference for checks outside loops and surfacing failures cleanly, curious whether you'd also want a `supports_hybrid` property on the ABC to make the fallback behavior explicit, rather than relying on `NotImplementedError` from `keyword_search_documents`.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Good question. Two reasons we rolled our own thin adapter instead of using `langchain_community.OpenSearchVectorSearch`:

1. **Pre-computed embeddings.** The deep-research ingestion pipeline pre-computes embeddings in bulk (chunking → embed → bulk insert in one pass). `langchain_community.OpenSearchVectorSearch` doesn't expose an `embeddings` kwarg on `add_texts` — it always calls the embedder internally, forcing a redundant round-trip. Our `_OpenSearchVectorStore.add_texts` accepts pre-computed vectors directly.

2. **Dependency boundary.** `langchain_oci` is a sibling package to `langchain_community`, not a consumer of it. Importing from community would create a circular-looking dependency in the monorepo and pull in community's transitive deps (boto3, google-cloud-*, etc.) for users who only want OCI integration.

The adapter is ~120 lines and only wraps what we actually use (`similarity_search_with_score`, `add_texts`, `delete`, `get_by_ids`). If the pre-computed-embeddings gap is closed upstream we can revisit.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Good call — done in 0446033.

Switched `_OpenSearchVectorStore.delete()` from a per-document loop to a single bulk request:

```python
bulk_body = [
    {"delete": {"_index": self._index_name, "_id": str(document_id)}}
    for document_id in ids
]
self._client.bulk(body=bulk_body, refresh=True)
```

One network round-trip regardless of how many IDs are passed.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Fair point — done in 0446033.

The happy path (no errors at all) is unchanged — returns `ids_list` directly. On partial failure we now log a `WARNING` for each failed document before filtering it out:

```python
logging.getLogger(__name__).warning(
    "OpenSearch bulk index failed id=%s reason=%s",
    ids_list[i],
    err.get("reason", err),
)
```

Callers already get back only the IDs that actually landed (the return-value contract from the earlier fix); now they also get a visible log line per failure instead of a silent drop.

libs/oci/langchain_oci/datastores/tools/search.py
fede-kamel: Implemented — replaced SearchTool + KeywordSearchTool with a single HybridSearchTool that runs RRF (Reciprocal Rank Fusion) over both semantic and keyword results. Factory now returns [stats, search, get_document]. Tested live on the PubMed OpenSearch index — hybrid consistently surfaces docs that either mode alone misses. Details in the PR comment above.

libs/oci/langchain_oci/datastores/vectorstores/opensearch.py
fede-kamel: Adding a third reason on top of the two above: **schema flexibility**.

`langchain_community.OpenSearchVectorSearch` hardcodes `text_field="text"` as the field it reads document content from (line ~485, 1065 in the community source). Our existing OpenSearch index stores text in `"content"`, not `"text"` — so community would silently read the wrong field and return empty page_content on every hit. Our adapter parameterises both `vector_field` and `search_fields` so it works against any index schema without reformatting the data.

Concrete proof — the PubMed index we're running against:
```
content    text         ← document text lives here
title      text
source     keyword
vector_field  knn_vector
```

`OpenSearchVectorSearch` would look for `_source["text"]` and get nothing. Our `_OpenSearchVectorStore` reads `_source["content"]` via `_extract_content()`, which also handles the `metadata.content` nesting some indices use.

YouNeedCryDear (approved):

PR #195, opened 2026-04-15
feat: Add datastore factory and package wiring

fede-kamel:
## Summary
- Add StoreSelector, create_datastore_tools factory, wire into langchain_oci package

## Dependency
**PR 5/13** of the deepagents agent stack. Depends on PR 3 + PR 4.

## Test plan
- [x] Unit tests pass

### DIFF
--- a/libs/oci/langchain_oci/__init__.py
+++ b/libs/oci/langchain_oci/__init__.py
@@ -1,7 +1,17 @@
 # Copyright (c) 2025 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
-from langchain_oci.agents.react import create_oci_agent
+from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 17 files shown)


fileames:
Sorry for the late review, some comments:
- The optional extra still allows langchain-oracledb>=1.2.0, but the ADB write path uses `OracleVS.add_documents(..., text_splitter=...)` when `chunk_on_write=True`. The floor should be bumped to at least langchain-oracledb>=1.3.0 for `text_splitter` support.
- I tested this in a clean env from the PR checkout and `numpy` was not installed transitively.   Command shape:
  `uv run python -c "import importlib.util; print(importlib.util.find_spec('numpy')); from langchain_oci import
  create_datastore_tools"` This printed `None` for `numpy` and then failed importing the new public export:
`ModuleNotFoundError: No module named 'numpy'`. Could you verify this on your side?

fileames (approved):

PR #196, opened 2026-04-15
test: Add datastore unit tests

fede-kamel:
## Summary
- Comprehensive unit tests for vectorstores, tools, selector, and factory (660 lines)

## Dependency
**PR 6/13** of the deepagents agent stack. Depends on PR 5.

## Test plan
- [x] 44 datastore tests pass

### DIFF
--- a/libs/oci/langchain_oci/__init__.py
+++ b/libs/oci/langchain_oci/__init__.py
@@ -1,7 +1,17 @@
 # Copyright (c) 2025 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
-from langchain_oci.agents.react import create_oci_agent
+from typing import TYPE_CHECKING, Any
+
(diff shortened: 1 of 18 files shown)


fileames (dismissed):

fileames:
- HybridSearchTool is now the actual factory-created search implementation, but it is not exported from `langchain_oci.datastores.tools / langchain_oci.datastores`. Direct import fails, while `SearchTool` and `KeywordSearchTool` are exported.
  - Question on StoreSelector semantics: should `default_store` be a true fallback when routing confidence is weak Right now `best_score` starts at `-1.0`, so the first store with any valid cosine score replaces `default_store`. That means `default_store` only affects routing in practice if every score is exactly `-1.0`, which is unlikely. If the intent is “always pick the highest cosine similarity,” the current behavior is fine, but then `default_store` is not really a fallback. If the intent is fallback behavior, should we keep the default when no store has positive / above-threshold similarity?

fileames (approved):

PR #197, opened 2026-04-15
feat: Add deepagents agent

fede-kamel:
## Summary
- Introduce create_deepagents_agent() with deepagents support and datastore tools

## Dependency
**PR 7/13** of the deepagents agent stack. Depends on PR 5.

## Test plan
- [x] Unit tests pass (76 total)

### DIFF
--- a/libs/oci/langchain_oci/__init__.py
+++ b/libs/oci/langchain_oci/__init__.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any
 from langchain_oci.agents.react.agent import create_oci_agent
 
 if TYPE_CHECKING:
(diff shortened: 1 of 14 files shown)


fileames:
Question on intended dependency behavior: should the datastore/lightweight path of `create_deepagents_agent` work without the `deepagents` package installed?

The implementation calls `_check_deepagents_prerequisites()` at the very start, so even calls that later route to `_build_lightweight()` still require `deepagents` to be installed. The tests say the datastore path “should avoid `deepagents` planning middleware” and assert `langchain.agents.create_agent` is used, but it still requires/inserts `deepagents`, so it does not clarify whether `deepagents` is meant to be optional for that path.

fileames (dismissed):

fileames (approved):

PR #199, opened 2026-04-15
test: Add agent integration test infrastructure

fede-kamel:
## Summary
- Shared conftest with ADB/OpenSearch/OCI fixtures and core agent integration tests

## Dependency
**PR 9/13** of the deepagents agent stack. Depends on PR 7.

## Test plan
- [x] 58 integration tests pass

### DIFF
--- a/libs/oci/Makefile
+++ b/libs/oci/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all format lint test tests integration_tests docker_tests help extended_tests
+.PHONY: all format lint test tests integration_tests docker_tests help extended_tests agents_integration_tests
 
 # Default target executed when no arguments are given to make.
 all: help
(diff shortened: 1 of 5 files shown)


fileames:
~~I think this new Makefile target needs one dependency fix. `agents_integration_tests` runs pytest with `--timeout=300`, but `pytest-timeout` is not declared in the project test/integration dependencies. In a clean env this target could fail with an unrecognized `--timeout` argument?~~
Looks like fixed on next PRs.

fileames (dismissed):

fede-kamel:
@fileames rebased onto latest `main`. After rebase, most of the original diff turned out to already be in main (carried in via the #197/#198 squash merges that pulled forward test files from this branch). The remaining delta is small and focused:

- `libs/oci/Makefile` — new `agents_integration_tests` target
- `libs/oci/tests/integration_tests/agents/conftest.py` — auto-bootstrap of the ADB test table + a few fixture improvements

Ready for another look when you have time — should be a quick review now.

fileames (approved):

PR #190, opened 2026-04-15
fix: Structured output and multi-turn tool call counting

fede-kamel:
## Summary
- Fix tool schema conversion to use `tool_call_schema` instead of `args_schema` to avoid serialization failures with injected runtime parameters (deepagents middleware)
- Add `supports_tool_choice` to GenericProvider so `with_structured_output` forces tool use
- Fix multi-turn tool call counting to scope to current turn only
- Add configurable request timeout via `auth.py`

## Test plan
- [x] Unit tests: `test_tool_schema_constraints` (45 tests), `test_tool_call_turn_scoping`
- [x] Integration tests: `test_tool_calling`, `test_response_format` (migrated from unit to integration)

**Split from #148** — independent fix, can merge in any order.

PR #201, opened 2026-04-15
test: Add deepagents integration tests

fede-kamel:
## Summary
- End-to-end deepagents tests with live ADB/OpenSearch and Gemini models

## Dependency
**PR 11/13** of the deepagents agent stack. Depends on PR 9 + #190.

## Test plan
- [x] 58 integration tests pass

### DIFF
--- a/libs/oci/langchain_oci/agents/common.py
+++ b/libs/oci/langchain_oci/agents/common.py
@@ -225,6 +225,6 @@ def _langgraph_schema_fallback() -> Iterator[None]:
     try:
         yield
     finally:
-        lg_pydantic.create_model = original  # type: ignore[assignment]
+        lg_pydantic.create_model = original
         for mod in rebound:
             mod.create_model = original  # type: ignore[attr-defined]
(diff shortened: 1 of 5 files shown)


fileames:
Thanks for adding the `OracleSaver` checkpointer test. One question: because it passes `middleware=[]`, `create_deepagents_agent` takes the lightweight `langchain.agents.create_agent` path rather than the full `deepagents.create_deep_agent` path. If the goal is to validate Oracle-backed checkpointing specifically with the full deepagents path, should there also be a variant that forces the deep path, for example by using a real deepagents option instead of `middleware=[]`?

fileames (dismissed):

fede-kamel:
@fileames rebased onto latest `main`. Two commits dropped as their content was already pulled into main via earlier squash merges (#197/#198). Remaining diff:

- `libs/oci/tests/integration_tests/agents/test_deepagents_integration.py` — new 1238-line deepagents integration suite (the substance of this PR)
- `libs/oci/tests/integration_tests/agents/conftest.py` — shared conftest improvements (same delta as #199)

Ready for re-review.

fileames (approved):

PR #209, opened 2026-05-02
feat(oci): parse inline <tool_call> blocks for DAC fine-tunes

### LINKED ISSUE #207: Qwen3 tool calls are not parsed into AIMessage.tool_calls, breaking DAC compatibility
When using Qwen3 models through langchain_oci.ChatOCIGenAI with LangChain tool calling, tool calls are not exposed through the standard LangChain AIMessage.tool_calls field.

The model appears to return a valid tool call in the message content using Qwen-style tags such as:

```
<tool_call>
{ ... }
</tool_call>
or:

<tool_calling>
{ ... }
</tool_calling>
```
However, langchain_oci currently leaves this serialized tool call inside AIMessage.content and returns an empty AIMessage.tool_calls list.

This makes Qwen3 models incompatible with DAC workflows that rely on LangChain’s standard tool-calling contract. DAC expects tool calls to be available in AIMessage.tool_calls, not embedded as raw text in the model response content.

**Expected behavior**

When a Qwen3 model returns a tool call, ChatOCIGenAI should parse the Qwen3 tool-call format and populate AIMessage.tool_calls with the corresponding LangChain tool call objects.

For example, after invoking a tool-bound model:

```
response = llm.bind_tools([WeatherQuery]).invoke(prompt)
response.tool_calls should contain the parsed tool call, including:
```

tool name
tool arguments
generated or provider-supplied tool call ID
The serialized <tool_call> / <tool_calling> block should either be removed from response.content or otherwise handled consistently with LangChain tool-calling behavior.

**Actual behavior**

response.tool_calls is empty.

The tool call is returned only as raw text inside response.content.

As a result, downstream LangChain/DAC orchestration cannot detect that a tool should be executed.

Impact

This prevents Qwen3 models served through OCI Generative AI from being used reliably in DAC scenarios that depend on tool calling.

The issue affects agentic flows where DAC expects LangChain-compatible tool-call metadata and does not parse provider/model-specific XML-like tool-call tags from the assistant content.

**Suggested fix**

Add Qwen3-specific post-processing in ChatOCIGenAI so that, when the model response contains a Qwen3 tool-call block, the integration extracts the JSON payload and converts it into LangChain tool_call objects.

The parser should support both observed tag variants:

```
<tool_call>...</tool_call>
<tool_calling>...</tool_calling>
It should parse payloads shaped like:

{
  "name": "ToolName",
  "arguments": {
    "arg1": "value"
  }
}
```
and also handle the case where arguments is itself a JSON-encoded string.

I will attach test case n.1 showing the current behavior.

fede-kamel:
## Summary

Closes #207.

Hermes / Llama / Qwen-style fine-tunes hosted on OCI Dedicated AI Cluster (DAC) endpoints emit tool calls inline as either `<tool_call>{...}</tool_call>` or `<tool_calling>{...}</tool_calling>` blocks inside the assistant message text instead of populating the OCI structured `tool_calls` field. On current `main`:

- `chat.invoke(...)` against a DAC fine-tune with `bind_tools(...)` returns `tool_calls=[]` and the raw XML lands in `content`.
- `chat.stream(...)` against the same DAC produces no `tool_call_chunks` at all — the raw XML just streams as text.

That makes any LangChain agent loop on a DAC fine-tune impossible without subclassing `ChatOCIGenAI` and overriding `_generate` (which doesn't help streaming anyway).

This PR adds the parsing inside `GenericProvider` so both paths converge — no subclassing, both tag variants supported.

### What changes

- **`chat_response_to_text` / `chat_response_to_text_from_dict`** strip both `<tool_call>...</tool_call>` and `<tool_calling>...</tool_calling>` blocks. Sync and async non-streaming paths share the parser.
- **`chat_tool_calls`** falls back to the XML extractor when the structured field is empty, returning small `_XmlToolCall` stand-ins shaped like `oci.generative_ai_inference.models.FunctionCall` (same attributes + `attribute_map`) so `OCIUtils.convert_oci_tool_call_to_langchain` works unchanged.
- **`chat_stream_to_text`** is now an incremental parser: appends each event delta to a per-stream buffer, drains complete blocks for either tag variant, and returns only the prefix that cannot still be the start of an opener. Partial `<` is held until disambiguated; once `<n` (or any non-`tool_call`/`tool_calling` second character) arrives, the held-back text is released.
- **`process_stream_tool_calls`** first emits XML blocks that completed during the current event as `tool_call_chunk(...)`, then falls through to the existing OCI structured-tool-call path.
- **`reset_stream_state` / `flush_stream_state`** are called from `_stream` and `_astream` at start / end-of-stream so a previous stream's leftover buffer can't leak forward, and trailing characters held back as a possible-opener prefix surface as a final delta if no opener arrives.
- **Mismatched open/close pairs** (`<tool_call>...</tool_calling>`) are intentionally NOT parsed and are left in the text so the issue stays visible to the caller rather than getting silently swallowed.
- **Argument shape:** the parser accepts both object-form `arguments` (most fine-tunes) and JSON-string-form `arguments` (some double-encode). The serialised string is forwarded to the existing downstream parser, which already handles both shapes.

The change is a strict superset: responses without inline markers flow through unchanged, native `tool_calls` still take precedence over the text fallback, malformed payloads are left in the text rather than silently dropped, and CohereProvider is untouched.

### Verification

- 42 unit tests in `tests/unit_tests/chat_models/test_xml_tool_calls.py` covering payload parsing, both tag variants, mixed variants in one response, mismatched open/close, the safe-emit split, the streaming buffer (blocks split across chunks for both tags, unclosed blocks at end-of-stream, partial-prefix safe-emit for both opener variants), and the chat-level fallback hooks.
- 15 live DAC integration tests in `tests/integration_tests/chat_models/test_dac_endpoint.py` (env-gated; skip cleanly without `DAC_ENDPOINT_OCID` / `DAC_COMPARTMENT_ID` / `DAC_SERVICE_ENDPOINT`). All pass against a live UK-region DAC fine-tune in ~12s.
- Pre-existing 302 unit tests still pass — Cohere/Gemini/Meta/OpenAI provider paths are untouched.

A 17-scenario end-to-end smoke run (sanitised) is published as a gist for visibility:
https://gist.github.com/fede-kamel/2506703bea52703cc4529a3993177730

## Test plan

- [x] `make lint` (ruff + ruff format + mypy `.`) clean on 3.12
- [x] `make test` — 344 passed, 12 skipped (302 pre-existing + 42 new XML parser tests)
- [x] Live DAC integration tests — 15/15 pass against a live UK-region fine-tune
- [x] Both `<tool_call>` and `<tool_calling>` tag variants verified end-to-end
- [x] CohereProvider / GeminiProvider / MetaProvider / OpenAIProvider regression-free (existing parametrized tests untouched)
- [ ] Reviewer to confirm against their own DAC endpoint if convenient

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -42,17 +42,25 @@ from pydantic import BaseModel
 from langchain_oci.chat_models.providers.base import Provider
 from langchain_oci.common.utils import OCIUtils
 
-# Hermes/Llama-style inline tool-call format. Some fine-tuned models hosted on
-# OCI Dedicated AI Clusters return tool calls as `<tool_call>{...}</tool_call>`
-# blocks inside the assistant message text instead of populating the structured
-# `tool_calls` field on the response. We extract those here so callers see
-# normal structured tool calls in both `_generate` and `_stream`.
-_XML_TOOL_OPEN = "<tool_call>"
-_XML_TOOL_CLOSE = "</tool_call>"
+# Hermes/Llama/Qwen-style inline tool-call format. Some fine-tuned models
+# hosted on OCI Dedicated AI Clusters return tool calls inside the assistant
+# message text using one of two equivalent tag pairs:
+#
+#   <tool_call>{"name": ..., "arguments": {...}}</tool_call>
+#   <tool_calling>{"name": ..., "arguments": {...}}</tool_calling>
+#
+# (Qwen3 has been observed emitting either form — see
+# https://github.com/oracle/langchain-oracle/issues/207.) We extract both
+# variants here so callers see normal structured tool calls in `_generate`
+# and `_stream`.
+_XML_TOOL_TAG_NAMES = ("tool_call", "tool_calling")
 _XML_TOOL_BLOCK_RE = re.compile(
-    re.escape(_XML_TOOL_OPEN) + r"\s*(.*?)\s*" + re.escape(_XML_TOOL_CLOSE),
+    r"<(?P<tag>" + "|".join(_XML_TOOL_TAG_NAMES) + r")>\s*(?P<body>.*?)\s*</(?P=tag)>",
     re.DOTALL,
 )
+# Used by the streaming buffer to detect that the trailing portion of the
+# buffer could still be the start of an opening tag for either variant.
+_XML_TOOL_OPENERS = tuple(f"<{name}>" for name in _XML_TOOL_TAG_NAMES)
 
 
 def _parse_xml_tool_call_payload(payload: str) -> Optional[Dict[str, Any]]:
@@ -83,20 +91,22 @@ def _parse_xml_tool_call_payload(payload: str) -> Optional[Dict[str, Any]]:
 
 
 def _extract_xml_tool_calls(text: str) -> Tuple[str, List[Dict[str, Any]]]:
-    """Strip ``<tool_call>...</tool_call>`` blocks and return parsed calls.
+    """Strip ``<tool_call>``/``<tool_calling>`` blocks and return parsed calls.
 
     Returns ``(cleaned_text, [{"id", "name", "arguments"}, ...])``. Blocks
     whose JSON payload doesn't parse are left intact in ``cleaned_text`` so
     no information is silently lost.
     """
-    if _XML_TOOL_OPEN not in text or _XML_TOOL_CLOSE not in text:
+    # Cheap pre-check: skip the regex scan entirely when the text doesn't
+    # contain either opener — the common case for plain assistant turns.
+    if not any(opener in text for opener in _XML_TOOL_OPENERS):
         return text, []
 
     parsed: List[Dict[str, Any]] = []
     cleaned_parts: List[str] = []
     cursor = 0
     for match in _XML_TOOL_BLOCK_RE.finditer(text):
-        payload = _parse_xml_tool_call_payload(match.group(1))
+        payload = _parse_xml_tool_call_payload(match.group("body"))
         if payload is None:
             # Malformed block — leave it in the text and continue scanning.
             continue
@@ -135,9 +145,9 @@ def _safe_emit_split(buffer: str) -> Tuple[str, str]:
     """Split a streaming buffer into (safe_to_emit, hold_back).
 
     ``hold_back`` covers any trailing portion that could still be the start of
-    a ``<tool_call>`` opening tag, or a complete-but-not-yet-closed tag whose
-    closing marker hasn't arrived. Everything before the earliest such ``<``
-    is safe to forward to the caller as text.
+    a ``<tool_call>`` / ``<tool_calling>`` opening tag, or a
+    complete-but-not-yet-closed tag whose closing marker hasn't arrived.
+    Everything before the earliest such ``<`` is safe to forward as text.
     """
     earliest_hold = len(buffer)
     pos = 0
@@ -146,9 +156,12 @@ def _safe_emit_split(buffer: str) -> Tuple[str, str]:
         if idx == -1:
             break
         rem = buffer[idx:]
-        # Either rem is a strict prefix of "<tool_call>" (still arriving) or
-        # rem already starts with "<tool_call>" (full opener, body in flight).
-        if _XML_TOOL_OPEN.startswith(rem) or rem.startswith(_XML_TOOL_OPEN):
+        # Either rem is a strict prefix of an opener (still arriving) or
+        # rem already starts with one (full opener, body in flight).
+        if any(
+            opener.startswith(rem) or rem.startswith(opener)
+            for opener in _XML_TOOL_OPENERS
+        ):
             earliest_hold = idx
             break
         pos = idx + 1
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -22,9 +22,10 @@ Currently, Google Gemini models have the broadest multimodal support on OCI.
 """
 
 import json
+import re
 import uuid
 import warnings
-from typing import Any, Callable, Dict, List, Literal, Optional, Type, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Type, Union
 
 from langchain_core.messages import (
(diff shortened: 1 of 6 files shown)


fede-kamel:
Closes #207.

Quick summary of what was found and what the fix does (full write-up in the PR body and on the issue):

| Symptom on `main` | Cause | Fix |
|---|---|---|
| `chat.invoke(...)` returns `tool_calls=[]` with raw `<tool_call>` XML in `content` | `GenericProvider.chat_response_to_text` returned the XML as-is; `chat_tool_calls` only looked at the structured `tool_calls` field, which Qwen3-style fine-tunes don't populate | `chat_response_to_text` / `..._from_dict` strip both `<tool_call>...</tool_call>` and `<tool_calling>...</tool_calling>` blocks; `chat_tool_calls` falls back to parsing them out of the text and synthesises `_XmlToolCall` objects shaped like `oci.generative_ai_inference.models.FunctionCall` so the existing `OCIUtils.convert_oci_tool_call_to_langchain` path keeps working |
| `chat.stream(...)` returns no `tool_call_chunks` and the raw XML streams as text | `GenericProvider.chat_stream_to_text` and `process_stream_tool_calls` operate per-event with no cross-event state, so a `<tool_call>...</tool_call>` block split across many events is invisible to them | `GenericProvider` now keeps a per-stream buffer (`_xml_stream_state`); `chat_stream_to_text` is incremental — it appends each delta, drains complete blocks, and returns only the safe-to-emit prefix. `process_stream_tool_calls` first emits any blocks that completed during the current event as `tool_call_chunk(...)`, then falls through to the existing structured-tool-call path. `_stream` / `_astream` call `reset_stream_state()` at start and `flush_stream_state()` at end so leftover buffer can't leak between streams |

Both tag variants from #207 (`<tool_call>` and `<tool_calling>`) are supported. Mismatched pairs (`<tool_call>...</tool_calling>`) are intentionally NOT parsed and are left in the text so the issue stays visible to the caller rather than getting silently swallowed.

**Verification artefacts:**
- 42 unit tests in `test_xml_tool_calls.py` (payload parsing, both variants, mixed variants in one response, mismatched pairs, safe-emit splits, streaming buffer including blocks split across chunks, partial-prefix safe-emit for both opener variants, chat-level fallbacks)
- 15 live DAC integration tests in `test_dac_endpoint.py` — env-gated; skip cleanly in CI
- Sanitised 17-scenario end-to-end smoke run: https://gist.github.com/fede-kamel/2506703bea52703cc4529a3993177730

cc @luigisaetta — happy to wire up your test case n.1 from #207 as an explicit regression test if you can paste the prompt + expected shape.

libs/oci/langchain_oci/chat_models/providers/generic.py
```diff
@@ -41,6 +42,131 @@
...
+_XML_TOOL_BLOCK_RE = re.compile(
+    r"<(?P<tag>" + "|".join(_XML_TOOL_TAG_NAMES) + r")>\s*(?P<body>.*?)\s*</(?P=tag)>",
+    re.DOTALL,
+)
+# Used by the streaming buffer to detect that the trailing portion of the
+# buffer could still be the start of an opening tag for either variant.
+_XML_TOOL_OPENERS = tuple(f"<{name}>" for name in _XML_TOOL_TAG_NAMES)
+
+
+def _parse_xml_tool_call_payload(payload: str) -> Optional[Dict[str, Any]]:
```
YouNeedCryDear: Those things should either be in utils or create another folder specifically for tool call parsing utils.

libs/oci/langchain_oci/chat_models/providers/generic.py
```diff
@@ -219,14 +363,68 @@ def chat_response_to_text_from_dict(self, response_data: Dict[str, Any]) -> str:
...
-            return ""
-        return "".join(part.get("text", "") for part in content if part.get("text"))
+            raw_delta = ""
+        else:
+            raw_delta = "".join(
+                part.get("text", "") for part in content if part.get("text")
+            )
+        return self._feed_xml_stream(raw_delta)
+
+    def reset_stream_state(self) -> None:
```
YouNeedCryDear, line 262: For those non-generic functions. Would it make sense to create a new QwenProvider?

YouNeedCryDear (changes requested):

fede-kamel:
@YouNeedCryDear thanks for the review.

**Re (1) — parser to utils:** agreed. Moved to `libs/oci/langchain_oci/common/xml_tool_call_parser.py` in 2aa6301. Sits next to the existing `OCIUtils` tool-call helpers in `common/`.

**Re (2) — `QwenProvider`:** I'd push back on the class split specifically, but I took the underlying point — most of the non-generic surface is now out of `GenericProvider` too. The new module owns both the pure parser (`extract_xml_tool_calls`, `parse_xml_tool_call_payload`, `safe_emit_split`, `XmlToolCall`) and the streaming state (`XmlStreamBuffer` with `feed` / `drain_completed` / `reset` / `flush`). `GenericProvider` is back to thin wiring around those, no per-stream buffer on `self`.

Three reasons `QwenProvider` would be the wrong shape:

1. **DAC dispatch can't see model family.** `_get_provider` in `llms/oci_generative_ai.py:170-196` routes any `model_id` starting with `CUSTOM_ENDPOINT_PREFIX` unconditionally to `"generic"` (line 186). The OCID is opaque — no served-model-family hint. A `QwenProvider` could only be reached by passing `provider=` explicitly, which regresses the UX issue #207 is fixing (agent loops on DAC fine-tunes "just working" without subclassing).
2. **The format is the Hermes function-calling convention (NousResearch), not Qwen-specific.** Qwen3 adopted it; Hermes-style fine-tunes of Llama use it too. Naming it `QwenProvider` would mislead Hermes/Llama users.
3. **The change is a strict superset for Generic.** Responses without inline markers flow through unchanged, native `tool_calls` still take precedence over the XML fallback. No regression surface for existing Generic users (Meta / Grok / OpenAI / Gemini / Mistral).

Happy to revisit splitting if DAC ever exposes a served-model-family hint, but a class split today forces opt-in for the very users this unblocks. Let me know if you want another pass on the layout.

The refactor is pure code-organization — no behavior change. 350 unit tests green (up from 344), 100% coverage on the new module, mypy + ruff clean over the whole tree.

YouNeedCryDear (approved):

PR #211, opened 2026-05-03
fix(oracledb): keep vector index in plan with VECTOR_INDEX_TRANSFORM hint (closes #130)

### LINKED ISSUE #130: JSON Search Index Not Used Together with Vector Index When Both Exist
Using a filter like the following, you can perform a search while filtering on metadata:

```
filter_dict = {"category": search_category}
vector_store.similarity_search(user_question, 8, filter=filter_dict)
```

Internally, this results in a JSON_EXISTS predicate being added to the WHERE clause, as shown below:

```
SELECT text, metadata,
       vector_distance(embedding, :embedding, COSINE) AS distance
FROM "ORAVS_DOCUMENTS"
WHERE JSON_EXISTS(
        metadata,
        '$.category?(@ == $val)'
        PASSING :value0 AS "val"
      )
ORDER BY distance
FETCH APPROX FIRST 8 ROWS ONLY;
```

For performance reasons, a JSON Search Index is typically created using the CREATE SEARCH INDEX statement.

With this query:

- If no JSON Search Index exists, the optimizer uses the Vector Index.
- If a JSON Search Index exists, only the JSON Search Index is used, and the Vector Index is not applied.

Therefore, it is necessary to add a hint to the query, such as:

```
SELECT /*+ VECTOR_INDEX_TRANSFORM(ORAVS_DOCUMENTS) */
       text, metadata,
       ...
```

to ensure the desired execution plan.

fede-kamel:
## Summary

Closes #130.

`_get_similarity_search_query` in `libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py` builds a `SELECT` that filters via `JSON_EXISTS(metadata, ...)` (when a metadata `filter` is passed) and orders by `vector_distance(...)`. As @TheKoguryo described in #130:

- Without a JSON Search Index on the table, the optimizer uses the vector index ✅
- With a JSON Search Index on the table (created via `CREATE SEARCH INDEX`), the optimizer picks the JSON Search Index for the `JSON_EXISTS` predicate and **drops the vector index from the plan entirely** — so the similarity search degrades to a full vector-distance computation across the JSON-filtered subset.

Adding the Oracle hint `/*+ VECTOR_INDEX_TRANSFORM({table_name}) */` after `SELECT` keeps the vector index in the plan when both indexes exist. With no JSON Search Index defined, the hint is a no-op, so the change is backwards-safe.

```diff
+    # `VECTOR_INDEX_TRANSFORM` keeps the vector index in the plan even when a
+    # JSON Search Index is also defined on the table. Without the hint, when
+    # both indexes exist the optimizer picks the JSON Search Index for the
+    # `JSON_EXISTS` filter and skips the vector index entirely, which kills
+    # similarity-search latency. See issue #130.
     query = f"""
-    SELECT
+    SELECT /*+ VECTOR_INDEX_TRANSFORM({table_name}) */
         text,
         metadata,
         ...
```

The hint references the same `{table_name}` identifier that's already interpolated into `FROM`, so callers who pass a quoted identifier get matching quoting in the hint and both refer to the same object.

## Credit

The original one-line fix was proposed by @TheKoguryo in #131; that PR was closed without merging because the OCA wasn't signed. This PR re-lands the same change and adds test coverage so the regression can't slip back in unnoticed.

## Tests

Three new unit tests in `tests/unit_tests/vectorstores/test_similarity_search_query.py`:

| Test | Asserts |
|---|---|
| `test_query_contains_vector_index_transform_hint_unfiltered` | hint present even without a filter |
| `test_query_contains_vector_index_transform_hint_with_filter` | hint present alongside `JSON_EXISTS` — the actual scenario from the bug |
| `test_hint_uses_caller_supplied_table_identifier` | quoted identifier flows through identically into both `FROM` and the hint |

## Verification

- `make lint` — ruff + ruff format + `mypy .` clean
- `make test` — 10 passed (7 pre-existing + 3 new), 1 skipped

## Test plan

- [x] `make lint`
- [x] `make test` on Python 3.12
- [x] CI matrix on 3.9 / 3.12 / 3.13 (will run on push)
- [ ] Reviewer to confirm: against an ADB with both a JSON Search Index and a vector index defined, similarity search with a metadata filter now uses the vector index in the execution plan

### DIFF
--- a/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
+++ b/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
@@ -927,8 +927,13 @@ def _get_similarity_search_query(
     if filter:
         where_clause = _generate_where_clause(filter, bind_variables)
 
+    # `VECTOR_INDEX_TRANSFORM` keeps the vector index in the plan even when a
+    # JSON Search Index is also defined on the table. Without the hint, when
+    # both indexes exist the optimizer picks the JSON Search Index for the
+    # `JSON_EXISTS` filter and skips the vector index entirely, which kills
+    # similarity-search latency. See issue #130.
     query = f"""
-    SELECT 
+    SELECT /*+ VECTOR_INDEX_TRANSFORM({table_name}) */
         text,
         metadata,
         vector_distance(embedding, :embedding,
--- a/libs/oracledb/tests/unit_tests/vectorstores/test_similarity_search_query.py
+++ b/libs/oracledb/tests/unit_tests/vectorstores/test_similarity_search_query.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2026 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Unit tests for the SQL emitted by ``_get_similarity_search_query``.
+
+The vector-index hint check (issue #130) keeps the vector index in the plan
+even when a JSON Search Index is also defined on the table — without the
+hint, the optimizer picks the JSON Search Index for the metadata filter and
+skips the vector index entirely, regressing similarity-search latency.
+"""
+
+from __future__ import annotations
+
+from langchain_community.vectorstores.utils import DistanceStrategy
+
+from langchain_oracledb.vectorstores.oraclevs import _get_similarity_search_query
+
+_TABLE = "ORAVS_DOCUMENTS"
+
+
+def test_query_contains_vector_index_transform_hint_unfiltered() -> None:
+    sql, _ = _get_similarity_search_query(
+        table_name=_TABLE,
+        distance_strategy=DistanceStrategy.COSINE,
+        k=8,
+    )
+    assert f"/*+ VECTOR_INDEX_TRANSFORM({_TABLE}) */" in sql
+
+
+def test_query_contains_vector_index_transform_hint_with_filter() -> None:
+    """The filter path is the actual scenario from issue #130 — once the
+    optimizer sees `JSON_EXISTS` and a JSON Search Index, it would otherwise
+    drop the vector index. The hint has to be present here too.
+    """
+    sql, _ = _get_similarity_search_query(
+        table_name=_TABLE,
+        distance_strategy=DistanceStrategy.COSINE,
+        k=8,
+        filter={"category": "research"},
+    )
+    assert f"/*+ VECTOR_INDEX_TRANSFORM({_TABLE}) */" in sql
+    assert "JSON_EXISTS" in sql
+
+
+def test_hint_uses_caller_supplied_table_identifier() -> None:
+    """Whatever identifier flows into ``FROM`` must also flow into the hint
+    so the two reference the same object — including any quoting the caller
+    chose to apply.
+    """
+    quoted = '"My Vector Table"'
+    sql, _ = _get_similarity_search_query(
+        table_name=quoted,
+        distance_strategy=DistanceStrategy.COSINE,
+        k=4,
+    )
+    assert f"/*+ VECTOR_INDEX_TRANSFORM({quoted}) */" in sql
+    assert f"FROM {quoted}" in sql


fede-kamel:
## Live optimizer verification

To prove the hint actually changes the optimizer's behavior (not just the SQL string), I created a test table on a real Oracle Autonomous DB with **both** a vector index and a JSON Search Index defined, then ran `EXPLAIN PLAN` on the OracleVS-shape query both without and with the hint.

### Setup

```sql
CREATE TABLE ISSUE130_TEST (
    id        RAW(16) DEFAULT SYS_GUID() PRIMARY KEY,
    text      CLOB,
    metadata  JSON,
    embedding VECTOR(16, FLOAT32)
);

-- 32 rows inserted, half with metadata.category='research'

CREATE VECTOR INDEX ISSUE130_TEST_VEC_IDX  ON ISSUE130_TEST (embedding)
  ORGANIZATION INMEMORY NEIGHBOR GRAPH DISTANCE COSINE WITH TARGET ACCURACY 95;

CREATE SEARCH INDEX ISSUE130_TEST_JSON_IDX ON ISSUE130_TEST (metadata) FOR JSON;
```

### Execution plan WITHOUT the hint (current `main`)

```
-----------------------------------------------------------------
| Id  | Operation                      | Name                   |
-----------------------------------------------------------------
|   0 | SELECT STATEMENT               |                        |
|*  1 |  COUNT STOPKEY                 |                        |
|   2 |   VIEW                         |                        |
|*  3 |    SORT ORDER BY STOPKEY       |                        |
|*  4 |     TABLE ACCESS BY INDEX ROWID| ISSUE130_TEST          |
|*  5 |      DOMAIN INDEX              | ISSUE130_TEST_JSON_IDX |
-----------------------------------------------------------------
Plan hash value: 2170752437
```

The optimizer picks the JSON Search Index for `JSON_EXISTS` and **drops the vector index from the plan entirely**. `vector_distance(...)` ends up being computed over every row matching the JSON filter — exactly the regression @TheKoguryo described.

### Execution plan WITH the hint (this PR)

```
----------------------------------------------------------------------------------------------------------------------
| Id  | Operation                                  | Name                                                            |
----------------------------------------------------------------------------------------------------------------------
|   0 | SELECT STATEMENT                           |                                                                 |
|*  1 |  COUNT STOPKEY                             |                                                                 |
|   2 |   VIEW                                     |                                                                 |
|*  3 |    SORT ORDER BY STOPKEY                   |                                                                 |
|*  4 |     TABLE ACCESS BY INDEX ROWID            | ISSUE130_TEST                                                   |
|   5 |      VECTOR INDEX HNSW SCAN PRE-FILTER     | ISSUE130_TEST_VEC_IDX                                           |
|   6 |       VIEW                                 | VW_HPJ_239B835B                                                 |
|   7 |        NESTED LOOPS OUTER                  |                                                                 |
|*  8 |         TABLE ACCESS BY INDEX ROWID        | ISSUE130_TEST                                                   |
|*  9 |          DOMAIN INDEX                      | ISSUE130_TEST_JSON_IDX                                          |
|  10 |         TABLE ACCESS BY INDEX ROWID BATCHED| VECTOR$ISSUE130_TEST_VEC_IDX$..._HNSW_ROWID_VID_MAP             |
|* 11 |          INDEX RANGE SCAN                  | VECTOR$ISSUE130_TEST_VEC_IDX$..._HNSW_ROWID_VID_IDX             |
----------------------------------------------------------------------------------------------------------------------
Plan hash value: 198100872
```

`VECTOR INDEX HNSW SCAN PRE-FILTER` is now at step 5 — the vector index is in the plan as a pre-filter, with the JSON Search Index narrowing the candidate set first via the nested-loop join. Different plan hash, materially different execution.

### Summary

| Run | Vector index in plan? | JSON Search Index in plan? | Plan hash |
|---|:---:|:---:|---|
| WITHOUT hint | ❌ | ✅ | 2170752437 |
| WITH hint    | ✅ `VECTOR INDEX HNSW SCAN PRE-FILTER` | ✅ | 198100872 |

The hint genuinely changes the optimizer's choice of plan when both indexes are present, restoring vector-index usage exactly as the issue requested. Test table was dropped after capture.

fede-kamel:
## Local verification

Ran the full test surface locally against a live Oracle 26ai ADB.

**Unit tests:** 10 passed, 1 skipped — the 3 new tests in `test_similarity_search_query.py` pass alongside the 7 existing.

**Lint:** `ruff check`, `ruff format --check`, and `mypy .` all clean on PR code.

**Integration tests (`tests/integration_tests/vectorstores/test_oraclevs.py`):** 22 passed, 11 failed in the full suite. Investigated each failure:

- **10 of 11 pass when run in isolation**, including every test that exercises the modified SQL path:
  - `test_perform_search_test` / `_async`
  - `test_db_filter_test` / `_async`
  - `test_from_texts_lobs` / `_async`
  - `test_filters` / `_async`
  - `test_add_texts_pool_test` / `_async`

  Suite-level failures are pre-existing `TBn` table-name collisions between sync/async pairs that share table names — orthogonal to this PR.

- **`test_add_texts_test_async` fails in isolation too.** Root cause is in the test itself at `test_oraclevs.py:1340-1343`, where two `asyncio.create_task` calls both run `OracleVS.acreate(..., "TB10", ...)` concurrently — whichever task loses the race hits `ORA-00955: name is already used by an existing object`. Verified the same test fails identically on current `main`, so it's a pre-existing test race, not introduced here.

Net: no regressions attributable to this PR.

## Nice-to-have (non-blocking)

1. **Rebase onto current `main`.** #215 landed after this branch was cut and introduced `_validate_index_type` / `_validate_int_param` in `oraclevs.py` plus env-var-driven creds in `test_oraclevs.py`. GitHub still reports `MERGEABLE` because there's no textual conflict, but a rebase keeps the diff honest and lets CI exercise the hint alongside the validator code path.
2. **Optional integration test for the original scenario.** A test that creates both a vector index and a `CREATE SEARCH INDEX … FOR JSON` on the same table, then runs `EXPLAIN PLAN FOR <similarity_search SQL>` and asserts the plan contains `VECTOR INDEX`, would automate the verification captured manually in the previous comment. Not blocking — the live plan-hash diff above is strong evidence on its own — but it would close the loop in CI for the next person who touches this code path.

fileames:
I think we should do this, one thing to consider is should we assume people will create JSON indices on their own or should we also add the code to create that index as well?

fede-kamel:
Thanks @fileames. My preference would be to **not** ship code to create the JSON Search Index, and leave that as the user's call. A few reasons:

- **It's not really our index.** A JSON Search Index lives on the `metadata` column for reasons that extend beyond similarity search — full-text JSON queries, `JSON_EXISTS`/`JSON_VALUE` from other apps on the same table, etc. The regression in #130 only manifests *when one already exists*; users who don't have one don't need one for similarity search to perform well — the vector index alone is fine.
- **The write-side cost is invisible to us.** A `CREATE SEARCH INDEX … FOR JSON` adds maintenance overhead on every `INSERT`/`UPDATE` to `metadata`. That's a tradeoff the user/DBA should own consciously, not a switch a vector-store library should flip on their behalf.
- **The hint already makes us robust to either choice.** With this PR, similarity search performs correctly whether or not a JSON Search Index is present, which is the right design — the library shouldn't push users toward a schema feature they may not need.
- **Asymmetry with `create_index()` is intentional.** The vector index is essential to OracleVS's reason for being; the JSON Search Index isn't. Wrapping its DDL would imply we're recommending it by default.

What I did do, to make the interaction discoverable for users who *do* have (or are considering) a JSON Search Index: added a `Note:` to the `similarity_search` / `asimilarity_search` docstrings explaining that the `VECTOR_INDEX_TRANSFORM` hint keeps the vector index in the plan when both indexes are present, and explicitly stating the library doesn't create the JSON index for you. That way `help(OracleVS.similarity_search)` surfaces the behavior without us taking on the schema decision.

Happy to revisit if you'd rather have an opt-in helper — but I think the docstring is the right level of guidance here.

fileames (approved):

PR #215, opened 2026-05-08
Add bug fixes to langchain-oracledb

fileames:
Syncs the updated `langchain-oracledb` package changes from the ADE integration tree into `libs/oracledb`.

  Key changes:
  - Harden Oracle SQL/DDL construction by validating and quoting identifiers.
  - Bind OracleDocLoader JSON parameters instead of interpolating them into SQL.
  - Validate text-search `k` values and returned columns before SQL generation.
  - Validate vector index parameters before DDL generation.
  - Clear Oracle session proxy state after embedding/vector-store operations.
  - Replace weak random object ID entropy in `OracleDocReader` with `secrets`.
  - Add/update unit and integration coverage for loader, retriever, vector-store, proxy cleanup, cache, and chat
  history behavior.

PR #220, opened 2026-05-11
Stabilize langchain-oracle integration tests and cleanup connection handling

diegoascencioqa:
This PR focuses on fixing and stabilizing the existing langchain-oracle integration tests so the full pytest run completes more reliably under the TPI harness.

Key changes:

- Fixed async and concurrent paths that were not terminating cleanly and could leave Oracle DB connections open.
- Prevented connection-resource exhaustion during full integration execution, which was contributing to unreliable or skipped paths.
- Moved OracleVS(...) creation out of the concurrent path so shared setup does not race during execution.
- Added explicit connection cleanup in fixture teardown so both sync and async connections are closed deterministically.
- Normalized credential handling in some tests so username and password are read from os.environ.
- Added missing os imports required for the environment-based credential lookup.
- Without the trailing comma, args=("Sri Ram") was a 7-char string (iterable), not a tuple, so the test wasn't actually concurrent before.

Net effect: the LangChain integration suite is more stable, with fewer avoidable skips and failures and more complete end-to-end execution.

### DIFF
--- a/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
+++ b/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
@@ -596,7 +596,7 @@ def test_create_hnsw_index_test() -> None:
         drop_index_if_exists(connection, "idx11")
     # With negative or out-of-bound values for all 4 of them, we get the same errors.
     # Expectation:Index not created
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
         vs = OracleVS(connection, model1, "TB14", DistanceStrategy.EUCLIDEAN_DISTANCE)
         create_index(
             connection,
@@ -879,7 +879,7 @@ async def test_create_hnsw_index_test_async() -> None:
         await adrop_index_if_exists(connection, "idx11")
     # with negative values/out-of-bound values for all 4 of them, we get the same errors
     # Expectation:Index not created
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
         vs = await OracleVS.acreate(
             connection, model1, "TB14", DistanceStrategy.EUCLIDEAN_DISTANCE
         )
--- a/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
+++ b/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
@@ -1193,13 +1193,9 @@ def test_add_texts_test() -> None:
 
     # PyTorch + Transformers are NOT thread-safe during initialization
     # Unitialize model creation OUTSIDE threads
-    model = HuggingFaceEmbeddings(
-        model_name="sentence-transformers/all-mpnet-base-v2"
-    )
+    model = HuggingFaceEmbeddings(model_name="sentence-transformers/all-mpnet-base-v2")
     # Create ONCE (this creates the table safely)
-    vs_obj = OracleVS(
-        connection, model, "TB10", DistanceStrategy.EUCLIDEAN_DISTANCE
-    )
+    vs_obj = OracleVS(connection, model, "TB10", DistanceStrategy.EUCLIDEAN_DISTANCE)
 
     # 6. Add 2 different record concurrently
     # Expectation:Successful
--- a/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
+++ b/libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
@@ -474,10 +474,10 @@ def test_create_hnsw_index_test() -> None:
     drop_table_purge(connection, "TB4")
 
     # 6. idx_type left empty
-    # Expectation: Rejected instead of falling back to untyped DDL construction
+    # Expectation:Index created
     vs = OracleVS(connection, model1, "TB5", DistanceStrategy.EUCLIDEAN_DISTANCE)
-    with pytest.raises(ValueError, match="idx_type must be HNSW"):
-        create_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
+    create_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
+    drop_index_if_exists(connection, "Hello")
     drop_table_purge(connection, "TB5")
 
     # 7. efconstruction passed as parameter but not neighbours
@@ -534,7 +534,7 @@ def test_create_hnsw_index_test() -> None:
     drop_index_if_exists(connection, "idx11")
     drop_table_purge(connection, "TB9")
     # index not created:
-    with pytest.raises(ValueError):
+    with pytest.raises(RuntimeError):
         vs = OracleVS(connection, model1, "TB10", DistanceStrategy.EUCLIDEAN_DISTANCE)
         create_index(
             connection,
@@ -549,7 +549,7 @@ def test_create_hnsw_index_test() -> None:
         )
         drop_index_if_exists(connection, "idx11")
     # index not created:
-    with pytest.raises(ValueError):
+    with pytest.raises(RuntimeError):
         vs = OracleVS(connection, model1, "TB11", DistanceStrategy.EUCLIDEAN_DISTANCE)
         create_index(
             connection,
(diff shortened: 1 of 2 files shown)


fileames (dismissed):

libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
```diff
@@ -1193,21 +1191,21 @@ def test_add_texts_test() -> None:
     vs_obj.add_texts(texts4, ids=ids8)
     drop_table_purge(connection, "TB9")
 
+    # PyTorch + Transformers are NOT thread-safe during initialization
+    # Unitialize model creation OUTSIDE threads
```
fede-kamel: typo: `Unitialize` -> `Initialize`

libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
```diff
@@ -1193,21 +1191,21 @@ def test_add_texts_test() -> None:
...
-        vs_obj = OracleVS(
-            connection, model, "TB10", DistanceStrategy.EUCLIDEAN_DISTANCE
-        )
         texts5 = [val]
         ids9 = texts5
         vs_obj.add_texts(texts5, ids=ids9)
 
-    thread_1 = threading.Thread(target=add, args=("Sri Ram"))
-    thread_2 = threading.Thread(target=add, args=("Sri Krishna"))
+    thread_1 = threading.Thread(target=add, args=("Sri Ram",))
```
fede-kamel, line 1208: Worth a line in the PR description: without the trailing comma, `args=("Sri Ram")` was a 7-char string (iterable), not a tuple, so the test wasn't actually concurrent before.

fede-kamel (commented):
Non-blocking since @fileames already approved. Questions inline on the assertion changes in `test_create_hnsw_index_test`.

PR body mentions normalized credential handling via `os.environ` and added `os` imports but those aren't in the visible diff for these two files. Follow-up commit?

Doing similar suite-stabilization on the oci side in #208 (Wikimedia 400s, gpt-oss-120b reasoning_content drift) if it's useful context. Running the integration suite locally against ADB before approving.

fede-kamel (commented):
Ran the changed tests against ADB. 2 fail, 54 pass, 48 skip:

```
FAILED tests/integration_tests/vectorstores/test_oraclevs.py::test_create_hnsw_index_test
FAILED tests/integration_tests/vectorstores/test_oraclevs.py::test_create_hnsw_index_test_async
```

Both at the `idx_type=""` block:

```
oraclevs.py:539: _validate_index_type(config, "HNSW")
oraclevs.py:226: raise ValueError(f"idx_type must be {expected_type}.")
ValueError: idx_type must be HNSW.
```

The validator was added in #215 (merged 2026-05-08). #220 was branched after #215 but the test changes assume pre-#215 behavior. Same logic for the four `ValueError -> RuntimeError` blocks below: `_validate_int_param` raises ValueError pre-DDL, so TB10/TB11/TB12/TB13 should stay ValueError. Only TB14 (`parallel="hello"`) is correctly RuntimeError: the str-vs-int `>` raises TypeError into the generic handler.

Not approving yet. Connection-teardown and threading fixes are fine on their own.

libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
```diff
@@ -474,10 +474,10 @@ def test_create_hnsw_index_test() -> None:
...
 
     # 6. idx_type left empty
-    # Expectation: Rejected instead of falling back to untyped DDL construction
+    # Expectation:Index created
     vs = OracleVS(connection, model1, "TB5", DistanceStrategy.EUCLIDEAN_DISTANCE)
-    with pytest.raises(ValueError, match="idx_type must be HNSW"):
-        create_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
+    create_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
+    drop_index_if_exists(connection, "Hello")
     drop_table_purge(connection, "TB5")
```
fede-kamel, line 481: Revert needed: `_validate_index_type` (added in #215, `oraclevs.py:220-227`) still rejects empty `idx_type`.

```suggestion
    # 6. idx_type left empty
    vs = OracleVS(connection, model1, "TB5", DistanceStrategy.EUCLIDEAN_DISTANCE)
    with pytest.raises(ValueError, match="idx_type must be HNSW"):
        create_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
    drop_table_purge(connection, "TB5")
```

fede-kamel (commented):
Suggestion blocks for the two failures inline. The 48 skipped tests in `test_standard.py` are the `VectorStoreIntegrationTests` base class pattern: Sync classes set `has_async=False`, Async classes set `has_sync=False`, so half the matrix skips by design.

libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
```diff
@@ -739,14 +739,12 @@ async def test_create_hnsw_index_test_async() -> None:
...
     vs = await OracleVS.acreate(
         connection, model1, "TB5", DistanceStrategy.EUCLIDEAN_DISTANCE
     )
-    with pytest.raises(ValueError, match="idx_type must be HNSW"):
-        await acreate_index(
-            connection, vs, params={"idx_name": "Hello", "idx_type": ""}
-        )
+    await acreate_index(connection, vs, params={"idx_name": "Hello", "idx_type": ""})
+    await adrop_index_if_exists(connection, "Hello")
     await adrop_table_purge(connection, "TB5")
```
fede-kamel, line 749: Same revert on the async mirror.

```suggestion
    # 6. idx_type left empty
    vs = await OracleVS.acreate(
        connection, model1, "TB5", DistanceStrategy.EUCLIDEAN_DISTANCE
    )
    with pytest.raises(ValueError, match="idx_type must be HNSW"):
        await acreate_index(
            connection, vs, params={"idx_name": "Hello", "idx_type": ""}
        )
    await adrop_table_purge(connection, "TB5")
```

libs/oracledb/tests/integration_tests/vectorstores/test_oraclevs.py
```diff
@@ -534,7 +534,7 @@ def test_create_hnsw_index_test() -> None:
     drop_index_if_exists(connection, "idx11")
     drop_table_purge(connection, "TB9")
     # index not created:
-    with pytest.raises(ValueError):
+    with pytest.raises(RuntimeError):
```
fede-kamel: TB10-TB13 (lines 537, 552, 567, 582, plus the async mirrors) should revert to `pytest.raises(ValueError)`. `_validate_int_param` at `oraclevs.py:540-542` raises ValueError pre-DDL for `efConstruction=0`, `neighbors=0`, `parallel=0`, `accuracy=120`.

Keep `RuntimeError` at line 599 and its async mirror: `parallel="hello"` raises TypeError in the `>` comparison, which falls through to the generic handler in `_handle_exceptions`.

fede-kamel (changes requested):

diegoascencioqa:
@fede-kamel  thank you so much for the detailed review and for taking the time to run the tests locally. I've addressed all the feedback. 
Please let me know if anything else needs attention!

fede-kamel:
Thanks for the quick turnaround @diegoascencioqa — everything else in this round looks great. The `idx_type` empty blocks, TB10–TB13 (sync + async), the `Initialize` typo fix, and the connection/teardown work from earlier are all good to go.

Quick follow-up that's on me: I missed one spot when I wrote the previous review. TB14 (`parallel="hello"`) actually goes through the same `_validate_int_param` from #215 that flips TB10–TB13 to `ValueError`. `_validate_vector_index_common` calls it at `oraclevs.py:233`, which raises `ValueError("parallel must be an integer.")` at `:213` *before* the DDL runs. So the two TB14 blocks should match what you already did for TB10–TB13:

```diff
- with pytest.raises(RuntimeError):
+ with pytest.raises(ValueError):
      vs = OracleVS(connection, model1, "TB14", DistanceStrategy.EUCLIDEAN_DISTANCE)
      create_index(
          connection,
          vs,
          params={
              ...
              "parallel": "hello",
              ...
          },
      )
```

Sync block at `test_oraclevs.py:599`, async at `:882`. The `U1.TB14` blocks in `test_add_texts_test` aren't affected — different scenario (schema-qualified name resolution failing at DDL).

Verified locally with both blocks switched: `test_create_hnsw_index_test` and `_async` both pass cleanly. Sorry for the back-and-forth on this one — once that lands, the PR is ready to go from my side.

diegoascencioqa:
@fede-kamel done, updated TB14 as well. Thanks again!

fede-kamel (approved):

PR #223, opened 2026-05-13
[Workflow] onboard Macaron GitHub Actions security checks

YouNeedCryDear:
## Summary
Add Macaron policy verification for GitHub Actions using the `check-github-actions` policy.

## What this change does
- Adds a dedicated workflow for Macaron verification
- Checks workflow definitions and local composite actions
- Uses pinned actions and disables persisted checkout credentials
- Pins existing external GitHub Actions references to full commit SHAs

## Why this is required
This helps detect unsafe or vulnerable GitHub Actions usage and reduces risk from software supply chain attacks targeting CI/CD pipelines.

## Developer impact
The new Macaron workflow runs for changes under `.github/workflows/**` and `.github/actions/**`, supports manual validation with `workflow_dispatch`, and runs weekly on schedule. Existing workflows should retain their current behavior while using pinned action references.

## Validation
- `actionlint`
- Ruby YAML parse for 9 workflow/action YAML files
- `git diff --check`
- Mutable external `uses:` reference scan

## Required follow-up
- Enable `Macaron policy verification` as a required status check for protected branches
- Resolve any policy findings before merge if applicable
- Update the Macaron rollout tracker after the workflow is merged, has run successfully, and the required check is enabled

### DIFF
--- a/.github/actions/poetry_setup/action.yml
+++ b/.github/actions/poetry_setup/action.yml
@@ -26,7 +26,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
       name: Setup python ${{ inputs.python-version }}
       id: setup-python
       with:
@@ -41,7 +41,7 @@ runs:
       # interpreter even when matrix.python-version is older.
       run: echo "POETRY_VIRTUALENVS_PREFER_ACTIVE_PYTHON=true" >> "$GITHUB_ENV"
 
-    - uses: actions/cache@v4
+    - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
       id: cache-bin-poetry
       name: Cache Poetry binary - Python ${{ inputs.python-version }}
       env:
@@ -94,7 +94,7 @@ runs:
       run: pipx install "poetry==$POETRY_VERSION" --python "/usr/bin/python3" --verbose
 
     - name: Restore pip and poetry cached dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
       env:
         SEGMENT_DOWNLOAD_TIMEOUT_MIN: "4"
         WORKDIR: ${{ inputs.working-directory == '' && '.' || inputs.working-directory }}
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
   workflow_dispatch:  # Allows manual triggering from GitHub UI
 
+permissions:
+  contents: read
+
 # If another push to the same PR or branch happens while this workflow is still running,
 # cancel the earlier run in favor of the next run.
 #
--- a/.github/workflows/_lint.yml
+++ b/.github/workflows/_lint.yml
@@ -32,7 +32,9 @@ jobs:
           - "3.9"
           - "3.12"
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
 
       - name: Free Disk Space
         run: |
--- a/.github/workflows/_lint.yml
+++ b/.github/workflows/_lint.yml
@@ -8,6 +8,9 @@ on:
         type: string
         description: "From which folder this pipeline executes"
 
+permissions:
+  contents: read
+
 env:
   POETRY_VERSION: "2.2.1"
   WORKDIR: ${{ inputs.working-directory == '' && '.' || inputs.working-directory }}
--- a/.github/workflows/_release.yml
+++ b/.github/workflows/_release.yml
@@ -27,7 +27,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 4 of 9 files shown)


fede-kamel (approved):
LGTM — straightforward CI supply-chain hardening.

Verified independently:
- All 9 pinned third-party action SHAs resolve to the version tags annotated next to them (`actions/checkout@v4` / `v6.0.2`, `actions/setup-python@v5`, `actions/cache@v4`, `actions/upload-artifact@v4`, `actions/download-artifact@v4`, `codespell-project/actions-codespell@v2`, `pypa/gh-action-pypi-publish@release/v1`, `oracle/macaron@v0.24.0`).
- Least-privilege `permissions: contents: read` added to every reusable workflow (and `contents: read` retained on the `_test_release.yml` publish job alongside its required `id-token: write`).
- `persist-credentials: false` added to every `actions/checkout` invocation.
- The new Macaron `check-github-actions` job passed on this PR, so the policy is self-consistent with the pinning changes.
- All CI matrix jobs pass on 3.9 / 3.12 / 3.13 for both libs.

Two tiny non-blocking observations (no action needed):
- `_release.yml` adds `cache-key: release` to the post-publish import job — fine, just noting it's a behavioral change unrelated to pinning.
- `_lint.yml` only difference beyond the security bits is a trailing-newline fix, which is harmless cleanup.

.github/actions/poetry_setup/action.yml
```diff
@@ -26,7 +26,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
```
fede-kamel, line 29: Spot-checked every third-party SHA in this PR against the upstream `git/ref/tags` API — all 9 resolve to the annotated tags exactly (`actions/checkout` v4 & v6.0.2, `actions/setup-python` v5, `actions/cache` v4, `actions/upload-artifact` v4, `actions/download-artifact` v4, `codespell-project/actions-codespell` v2, `pypa/gh-action-pypi-publish` release/v1, `oracle/macaron` v0.24.0). No typo-squat risk.

.github/workflows/CI.yml
```diff
@@ -7,6 +7,9 @@ on:
   pull_request:
   workflow_dispatch:  # Allows manual triggering from GitHub UI
 
+permissions:
+  contents: read
+
```
fede-kamel, line 12: Top-level `permissions: contents: read` here plus on every reusable workflow matches what the jobs actually do (read code, run tests, publish via OIDC). No workflow writes back to the repo, so least-privilege is safe to apply across the board.

.github/workflows/_lint.yml
```diff
@@ -69,4 +74,4 @@ jobs:
       - name: Run linting
         working-directory: ${{ inputs.working-directory }}
         run: |
-          make lint
\ No newline at end of file
+          make lint
```
fede-kamel, line 77: Non-blocking: trailing-newline fix is unrelated to Macaron / pinning. Harmless to bundle but worth mentioning so reviewers don't go hunting for a semantic change.

.github/workflows/_release.yml
```diff
@@ -139,14 +146,17 @@ jobs:
...
+        with:
+          persist-credentials: false
 
       - name: Set up Python + Poetry ${{ env.POETRY_VERSION }}
         uses: "./.github/actions/poetry_setup"
         with:
           python-version: ${{ env.PYTHON_VERSION }}
           poetry-version: ${{ env.POETRY_VERSION }}
           working-directory: ${{ inputs.working-directory }}
+          cache-key: release
```
fede-kamel, line 159: Non-blocking observation: this is the one line in the PR that changes runtime behavior beyond SHA pinning + permissions. Adding `cache-key: release` isolates the import-published-package job into its own Poetry cache bucket so it doesn't collide with build/test caches. Worst case on first run after merge: one cold-cache miss. Fine.

.github/workflows/_test_release.yml
```diff
@@ -66,6 +71,7 @@ jobs:
       - build
     runs-on: ubuntu-latest
     permissions:
+      contents: read
```
fede-kamel, line 74: Good — `contents: read` made explicit alongside the pre-existing `id-token: write` (needed for PyPI trusted publishing). Keeps the least-privilege intent obvious and avoids GitHub's implicit-default behavior.

.github/workflows/macaron-check-github-actions.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run Macaron check-github-actions policy
+        uses: oracle/macaron@4ddb55e3c9ef2c77b548be55c557078c4476fd9c # v0.24.0
+        with:
+          repo_path: ./
+          policy_file: check-github-actions
```
fede-kamel, line 36: The Macaron `check-github-actions` job passed on this PR (job [75897536615](https://github.com/oracle/langchain-oracle/actions/runs/25831594485/job/75897536615)) — so the new policy is self-consistent with the pinning the PR itself introduces. Once this lands, recommend wiring it as a required status check for `main` per the PR description's follow-up list.

fede-kamel (commented):
Inline notes from my review — main review state was already submitted as APPROVED.

PR #224, opened 2026-05-14
Fix Oracle checkpoint BLOB writes for empty and large bytes

fileames:
## Summary

Fix Oracle checkpoint BLOB persistence for empty and large byte payloads.

`oracledb` async `executemany()` can bind raw Python `bytes` incorrectly for BLOB columns in the checkpoint `MERGE`
paths. Empty bytes can round-trip as `NULL`, and larger byte payloads can fail with `ORA-01461`. This affects both
pending writes and checkpoint blob storage.

This change converts byte payloads to temporary BLOB LOBs before `executemany()` and normalizes empty BLOB readback to `b""` before typed deserialization.

## Changes

- Added shared sync/async BLOB bind helpers for checkpoint writes.
- Updated `OracleSaver` and `AsyncOracleSaver` checkpoint BLOB paths to use temporary LOB binds.
- Preserved zero-length BLOB round-trips during checkpoint and pending-write loads.
- Added sync/async regression coverage for empty bytes and 32KB byte payloads.

### DIFF
--- a/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/_lob.py
+++ b/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/_lob.py
@@ -1,0 +1,38 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+import oracledb
+
+
+def with_blob_lobs(
+    conn: oracledb.Connection, params: Sequence[dict[str, Any]]
+) -> list[dict[str, Any]]:
+    """Replace byte BLOB bind values with temporary LOBs for executemany."""
+    lob_params = []
+    for param in params:
+        blob = param.get("blob")
+        if isinstance(blob, (bytes, bytearray)):
+            lob = conn.createlob(oracledb.DB_TYPE_BLOB)
+            if blob:
+                lob.write(bytes(blob))
+            param = {**param, "blob": lob}
+        lob_params.append(param)
+    return lob_params
+
+
+async def awith_blob_lobs(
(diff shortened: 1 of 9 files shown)


RichmondAlake (approved):

fede-kamel:
@fileames taking a look

fede-kamel (approved):
Pulled the branch and ran the new tests against `origin/main` to confirm the bug. 8 cases, 8 failures, exactly matching the two modes in the description:

| Path | Column | Failure on main |
|---|---|---|
| `aput_writes` / `put_writes` → `checkpoint_writes.blob` | `BLOB NOT NULL` (`base.py:59`) | `ORA-01400: cannot insert NULL into ("CHECKPOINT_WRITES"."BLOB")` |
| `aput` / `put` → `checkpoint_blobs.blob` | `BLOB` nullable (`base.py:47`) | `AssertionError: assert None == b''` — `b""` written then read back as `None` |

With the PR applied, all 8 cases pass. Routing through `conn.createlob(DB_TYPE_BLOB)` sidesteps oracledb's `b"" → NULL` coercion, and `_load_typed_blob` (`base.py:314-317`) closes the read side so the round-trip is symmetric. The sync/async split in `_lob.py` is necessary — `AsyncConnection.createlob` returns a coroutine.

One non-blocking thought: the temp LOBs aren't explicitly `lob.close()`'d. Python GC will free them when `lob_params` falls out of scope and the impl releases temp segments on cursor/connection close, but on long-lived pool connections doing many checkpoints it's worth keeping an eye on `V$TEMPORARY_LOBS` if you ever see session-temp pressure. Doesn't need to change in this PR.

LGTM.

PR #225, opened 2026-05-15
Fix OCI Generic provider structured JSON schema handling

varunyn:
## Summary

Fixes OCI Generic provider handling for LangChain structured-output schemas created as raw JSON Schema dictionaries.

This specifically unblocks LangChain middleware such as [LLMToolSelectorMiddleware](https://reference.langchain.com/python/langchain/agents/middleware/tool_selection/LLMToolSelectorMiddleware), which builds an internal JSON Schema dict and calls `model.with_structured_output(schema)` without passing `method="json_schema"`.

## Changes

- Allow `GenericProvider.convert_to_oci_tool()` to accept JSON Schema dicts, matching the behavior already supported by the Cohere provider path.
- Route Generic dict, Pydantic, callable, and BaseTool schemas through the same schema normalization pipeline:
  - resolve `$ref` / `$defs`
  - resolve `anyOf`
  - sanitize OCI-incompatible schema fields
- Extend Cohere dict and Pydantic/callable conversion paths to use the same sanitizer.
- Update schema sanitization to:
  - remove `x-*` extension keys recursively
  - lower string `const` values to `enum: ["value"]`
  - strip non-string `const` values to avoid OCI enum type errors
- Add regression tests for Generic structured-output dict schemas and sanitizer behavior.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/cohere.py
+++ b/libs/oci/langchain_oci/chat_models/providers/cohere.py
@@ -665,6 +665,9 @@ class CohereProvider(Provider):
                     "Unsupported dict type. Tool must be a BaseTool instance, "
                     "JSON schema dict, or Pydantic model."
                 )
+            schema = OCIUtils.resolve_schema_refs(tool)
+            schema = OCIUtils.resolve_anyof(schema)
+            schema = OCIUtils.sanitize_schema(schema)
             return self.oci_tool(
                 name=tool.get("title"),
                 description=tool.get("description"),
@@ -679,12 +682,15 @@ class CohereProvider(Provider):
                         ),
                         is_required="default" not in p_def,
                     )
-                    for p_name, p_def in tool.get("properties", {}).items()
+                    for p_name, p_def in schema.get("properties", {}).items()
                 },
             )
         elif (isinstance(tool, type) and issubclass(tool, BaseModel)) or callable(tool):
             as_json_schema_function = convert_to_openai_function(tool)
             parameters = as_json_schema_function.get("parameters", {})
+            parameters = OCIUtils.resolve_schema_refs(parameters)
+            parameters = OCIUtils.resolve_anyof(parameters)
+            parameters = OCIUtils.sanitize_schema(parameters)
             properties = parameters.get("properties", {})
             fn_name = as_json_schema_function.get("name", "")
             return self.oci_tool(
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -778,12 +778,11 @@ class GenericProvider(Provider):
         self,
         tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
     ) -> Dict[str, Any]:
-        """Convert a BaseTool instance, TypedDict or BaseModel type
-        to a OCI tool in Meta's format.
+        """Convert a tool definition to an OCI tool in Meta's format.
 
         Args:
-            tool: The tool to convert, can be a BaseTool instance, TypedDict,
-                or BaseModel type.
+            tool: The tool to convert, can be a BaseTool instance, JSON schema
+                dictionary, TypedDict, callable, or BaseModel type.
 
         Returns:
             Dict containing the tool definition in Meta's format.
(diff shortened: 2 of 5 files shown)


fede-kamel (changes requested):
## Review

Nice work covering the JSON-schema-dict path for `LLMToolSelectorMiddleware` and the related cases. **However, this PR introduces a regression on the Pydantic structured-output path**, and unit tests don't catch it because all the regressions surface as integration failures.

### Regression: `test_structured_output_no_docstring` fails for every model

On `main`, `tests/integration_tests/chat_models/test_response_format.py::test_structured_output_no_docstring` passes against `meta.llama-3.3-70b-instruct` and `google.gemini-2.5-flash`. Re-running the exact same test against this PR fails on **all 8 parametrized models** (Llama 3.3 / Gemini 2.5 Flash / Llama-4 Scout / Grok-4 fast / GPT-5.1 / GPT-5.2 / GPT-4.1 / Cohere command-a):

```
8 failed, 58 passed, 684 warnings in 297.10s
FAILED ...test_structured_output_no_docstring[meta.llama-3.3-70b-instruct]
FAILED ...test_structured_output_no_docstring[google.gemini-2.5-flash]
...
```

### Root cause

The Pydantic-to-OCI path now routes through `OCIUtils.sanitize_schema()` (new in this PR):

```python
if (isinstance(tool, type) and issubclass(tool, BaseModel)) or callable(tool):
    as_json_schema_function = convert_to_openai_function(tool)
    parameters = as_json_schema_function.get("parameters", {})
    resolved_params = OCIUtils.resolve_schema_refs(parameters)
    resolved_params = OCIUtils.resolve_anyof(resolved_params)
    resolved_params = OCIUtils.sanitize_schema(resolved_params)   # ← new
```

But `sanitize_schema` (existing code in `common/utils.py`) recurses into nested dicts and drops every key whose name is `"title"`:

```python
for key, value in schema.items():
    if key == "title":
        continue
```

That was originally intended to drop the JSON-Schema metadata `title` at the top level (a sibling of `type` / `description`). Because `sanitize_schema` recurses into `properties`, it also strips any **field literally named `title`**.

The test's schema has exactly such a field:

```python
class BugReport(BaseModel):
    title: str = Field(description="Short bug title")
    severity: str
    steps_to_reproduce: str
```

What the OCI request looked like before this PR vs. after, for the same Pydantic class:

```diff
 "properties": {
-  "title":               {"type": "string", "description": "Short bug title"},
   "severity":            {"type": "string", "description": "..."},
   "steps_to_reproduce":  {"type": "string", "description": "..."}
 },
 "required": ["title", "severity", "steps_to_reproduce"]
```

So:
1. The LLM never sees a `title` parameter in the tool definition.
2. The model returns a `BugReport` without `title`.
3. `required` still lists `title`, so Pydantic raises `ValidationError`.

### Suggested fix

Make the `"title"`-key skip apply only at the top level, where JSON-Schema metadata lives — never inside `properties`. Two options:

**Option A — top-level only:**

```python
@staticmethod
def sanitize_schema(schema: Any, *, _depth: int = 0) -> Any:
    if not isinstance(schema, dict):
        return schema
    sanitized: Dict[str, Any] = {}
    for key, value in schema.items():
        if _depth == 0 and key == "title":
            continue
        ...
        sanitized[key] = OCIUtils.sanitize_schema(value, _depth=_depth + 1)
```

**Option B — sanitize the top-level dict only when called, recurse explicitly into `properties` / `items` / `anyOf` etc.** (cleaner but a bigger refactor).

A regression test that mirrors `test_structured_output_no_docstring` but at the unit-test level (no network) would also be helpful — pass a Pydantic model that has a field literally named `title` through `convert_to_oci_tool` and assert the field survives.

### Everything else looks good

- Unit tests: **355 pass / 12 skipped** locally — your 5 new tests all pass.
- Lint clean (ruff + format).
- The `convert_to_oci_tool` dict branch and the `x-*` / non-string `const` sanitizer additions are well-targeted and the rationale (LLMToolSelectorMiddleware) is clearly documented.

Happy to re-test once the `title` recursion bug is fixed.

fede-kamel (approved):
Pushed the `title` / `const` / `x-*` recursion fix to your branch (commit efd7f67) — `sanitize_schema` now exempts the keys inside `properties`/`$defs`/`definitions` (user-defined names) from the metadata-key drop, while still stripping the same keys as metadata when they appear as siblings of `type`/`description`. Two new unit tests cover the regression. All 8 previously-failing `test_structured_output_no_docstring[*]` cases now pass against live OCI GenAI.

fede-kamel:
## Update — fix pushed, CI green

Pushed commit [`efd7f67`](../commit/efd7f67) on top of your branch to address the `title`-recursion regression I flagged in the previous review. Summary of what changed and why:

### The bug

`OCIUtils.sanitize_schema` recursively iterated every dict in a schema and dropped any key called `title` / `const` / `x-*`. That was correct when those keys appear as JSON-Schema *metadata* (siblings of `type` / `description`), but `properties` / `$defs` / `definitions` are dicts whose keys are *user-defined names*. The same drop was wiping out fields literally called `title` (or `const`, or `x-…`).

Concretely, the test's Pydantic model:

```python
class BugReport(BaseModel):
    title: str = Field(description="Short bug title")
    severity: str
    steps_to_reproduce: str
```

was being sent to OCI as:

```json
{
  "properties": {
    "severity": { ... },
    "steps_to_reproduce": { ... }
  },
  "required": ["severity", "steps_to_reproduce"]
}
```

— `title` silently dropped from both `properties` and `required`. The LLM filled what it saw; Pydantic on the way back then raised `ValidationError` because `title` is still required in the Python class. This regressed all 8 parametrized models of `test_structured_output_no_docstring`.

### The fix (`libs/oci/langchain_oci/common/utils.py`)

When `sanitize_schema` encounters a key called `properties`, `$defs`, or `definitions` and its value is a dict, recurse into each value as an independent schema but **don't apply the metadata-key drops to that dict's keys**. Metadata `title` / `const` / `x-*` *inside* a property's value still get stripped as before.

### Regression tests (`tests/unit_tests/chat_models/test_tool_schema_constraints.py`)

Two new unit tests, both passing:

- `test_sanitize_schema_preserves_user_fields_named_title` — schema with fields literally named `title`, `const`, `x-flag`; asserts they survive sanitisation AND that metadata `title` inside a property's value is still stripped.
- `test_sanitize_schema_preserves_user_defined_definition_names` — same guarantee for `$defs` / `definitions`.

### Verification

- 357 unit tests pass (+2 new), ruff/format clean, mypy clean.
- All 8 previously-failing `test_structured_output_no_docstring[*]` integration tests now pass against the live OCI Generative AI endpoint (Llama 3.3, Gemini 2.5 Flash, Llama 4 Scout, Grok-4 fast, GPT-5.1, GPT-5.2, GPT-4.1, Cohere command-a) in ~20s total.
- All 11 CI checks on the PR now green.

The original intent of your PR — handling JSON-schema dicts for `LLMToolSelectorMiddleware`, lifting the Cohere sanitizer to the Generic dict/Pydantic path, and the `x-*` / non-string `const` cleanup — is unchanged. The fix only narrows the recursion semantics inside `sanitize_schema` so user-defined names survive.

Happy to revise if you'd prefer a different approach (depth tracker, or only special-casing the top-level call rather than the `properties` recursion). Otherwise, this should be ready to merge.

fede-kamel:
Merged 🎉 Thanks @varunyn — really nice piece of work. Lifting the schema sanitization to the Generic + Pydantic + Cohere dict paths is a clean unblock for `LLMToolSelectorMiddleware` (and any other middleware that goes through `with_structured_output(<dict>)`), and the `x-*` / non-string `const` cleanup in the sanitizer is going to save people other obscure failures down the line.

The 5 unit tests you wrote made the regression I introduced (the recursive `title`-drop) easy to diagnose and patch around without needing to throw out any of your work. Appreciated.

PR #231, opened 2026-06-02
Handle async saver misuse and store schema consistency

fileames:
This PR improves LangGraph Oracle integration behavior around async usage and store setup recovery.

- Adds explicit async method implementations on the synchronous `OracleSaver` that raise a clear `NotImplementedError`,
directing users to `AsyncOracleSaver` when using LangGraph async APIs.
- Adds schema consistency checks for `OracleStore.setup()` and `AsyncOracleStore.setup()` so setup fails clearly when
migration tables indicate prior setup but required backing tables are missing.
- Covers both sync and async paths with focused tests for checkpoint async misuse and inconsistent store/vector schema
state.

## Testing

- Added `tests/test_checkpoint_sync_async_errors.py`
- Added `tests/test_store_setup_schema_consistency.py`

### DIFF
--- a/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/sync.py
+++ b/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/sync.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import threading
 import time
 from collections import defaultdict
-from collections.abc import Iterator, Sequence
+from collections.abc import AsyncIterator, Iterator, Sequence
 from contextlib import contextmanager
 from typing import Any
 
@@ -25,6 +25,11 @@ from langgraph_oracledb.checkpoint.oracle._lob import with_blob_lobs
 from langgraph_oracledb.checkpoint.oracle.base import BaseOracleSaver
 
 Conn = _internal.Conn  # For backward compatibility
+
+_AIO_ERROR_MSG = (
+    "OracleSaver is synchronous and does not support async checkpoint operations. "
+    "Use AsyncOracleSaver with LangGraph async APIs such as graph.ainvoke(...)."
+)
 
 
 def _validate_conn_string(conn_string: str):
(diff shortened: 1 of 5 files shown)


RichmondAlake (approved):

fede-kamel (approved):
## Review — LGTM, validated end-to-end ✅

Reviewed the diff and validated it locally with **unit tests + integration tests against a live free-tier Oracle Autonomous Database (23ai)**. The async-misuse errors and the schema-consistency guard both behave exactly as intended, on the sync and async paths.

### Correctness

- **Async stubs on `OracleSaver`.** `BaseCheckpointSaver` already raises a bare `NotImplementedError` for `aget_tuple`/`alist`/`aput`/`aput_writes`/`adelete_thread`, so this is a pure message-quality improvement (no behavior change) that points users to `AsyncOracleSaver`. The `alist` async-generator idiom (`raise` followed by an unreachable `yield`) is correct — it raises on iteration.
- **Schema-consistency guard.** Traced the migration indices: `store` is created at migration `0`, `store_configs` at `3`, `store_vectors` at vector-migration `0` — so the `>= 0` / `>= 3` thresholds are accurate. `_get_version` returns `-1` on a fresh table, so the guard correctly **skips brand-new databases** (no false positive on first `setup()`) and only fires when a migration version is recorded but the backing table is missing.

### What I tested

**Unit (mocked) — 9/9 pass:** `tests/test_checkpoint_sync_async_errors.py`, `tests/test_store_setup_schema_consistency.py`.

**Integration (live free-tier Oracle 23ai, real SQL, no mocks) — 8/8 pass:**

| Check | Result |
|---|---|
| sync fresh `setup()` — no false positive on a clean DB | ✅ |
| sync idempotent re-`setup()` on a healthy store | ✅ |
| sync `setup()` raises on missing **store** table | ✅ `…STORE_MIGRATIONS_… records migration 4, but STORE_… is missing…` |
| sync vector-store fresh `setup()` | ✅ |
| sync `setup()` raises on missing **store_vectors** table | ✅ `…VECTOR_MIGRATIONS_… records migration 1, but STORE_VECTORS_… is missing…` |
| sync `OracleSaver` async methods → `NotImplementedError("…Use AsyncOracleSaver…")` | ✅ |
| async fresh `setup()` — no false positive | ✅ |
| async `setup()` raises on missing store table | ✅ |

All test tables were torn down afterward (verified zero residue).

### Suggestions (non-blocking)

1. **Add `libs/langgraph-oracledb` to CI.** The `CI.yml` matrix is currently only `libs/oci` and `libs/oracledb`, so this library — and these new tests — don't run in CI; I validated everything locally. Strongly recommend adding it (lint + test) as a companion change, otherwise these tests aren't exercised on future PRs.
2. **De-magic `version >= 3`.** Correct today, but it's coupled to the `MIGRATIONS` list order. A named constant (e.g. `_STORE_CONFIGS_MIGRATION_VERSION = 3`) or a short comment would prevent a silent break if migrations are ever reordered.

The behavior is correct and verified. Approving — please consider the CI addition alongside this.

PR #233, opened 2026-06-04
fix(langchain-oracledb): keep vector index in plan with VECTOR_INDEX_TRANSFORM hint

sudarshan12s:
The fix is similar to python vectorstore https://github.com/oracle/langchain-oracle/pull/211

Along with the fix, 
minor test changes are done to let the suite run complely.


sudarshan12s:
@fileames, @fede-kamel, could you review this one? It's similar to the Python fix in PR #211.

libs/js/langchain-oracledb/src/vectorstores.ts
```diff
@@ -927,8 +927,13 @@ export class OracleVS extends VectorStore {
     try {
       const bindValues: unknown[] = [this.prepareQueryVector(query)];
 
+      // VECTOR_INDEX_TRANSFORM keeps the vector index in the plan even when a
+      // JSON Search Index is also defined on the table. Without the hint, when
+      // both indexes exist, the optimizer can pick the JSON Search Index for a
+      // JSON_EXISTS filter and skip the vector index, hurting search latency.
       let sqlQuery = `
-      SELECT external_id,
+      SELECT /*+ VECTOR_INDEX_TRANSFORM(${this.tableName}) */
```
fede-kamel: **The hint can change results, not just the plan, when a filter is present.** With the hint, the optimizer runs the approximate top-k through the vector index *first* and applies the `JSON_EXISTS` filter *after* (post-filtering). If the nearest neighbors don't match the filter, the query returns fewer than k rows — possibly zero — and `similaritySearchByVectorReturningEmbeddings` then throws `No rows found.`

Repro on 26ai (IVF index + JSON search index, 12 rows, 6 matching `category == "books"`, k=3, query embedding closest to the non-matching rows):

```
WITH hint    -> rows: 0
WITHOUT hint -> rows: 3   (correct top-3 from the 6 matching rows)
```

Without the hint the optimizer pre-filters via the JSON domain index and the results are exact. So as written, this trades correctness/recall on filtered searches for latency, and in the worst case turns a valid query into an exception.

Suggestions: apply the hint only when no filter is present, or over-fetch (k × factor) and post-filter client-side, or at minimum document the approximate semantics and stop throwing on empty results. Whatever is decided should also be revisited in the Python fix (oracle/langchain-oracle#211), which has the same behavior by construction.

(The tests in this PR can't catch this: the EXPLAIN PLAN test never fetches rows, and the filter tests use k=4 with only 4 docs, so the approximate top-k always covers the whole table.)

libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts
```diff
@@ -1350,6 +1358,102 @@ describe("OracleVectorStore", () => {
...
+          text,
+          metadata,
+          vector_distance(embedding, :embedding, ${oraclevs.distanceStrategy}) as distance,
+          embedding
+        FROM ${oraclevs.tableName}
+        WHERE JSON_EXISTS(metadata, '$.category?(@ == "books")')
+        ORDER BY distance
+        FETCH APPROX FIRST 4 ROWS ONLY
+        `,
+        { embedding: queryVector }
```
fede-kamel: **This breaks `pnpm build`** (and therefore `prepack`/release): binding a `Float32Array` in a named-binds object doesn't typecheck — `@types/oracledb`'s `BindParameters` record doesn't include typed arrays, so no `execute()` overload matches (TS2769 at the template literal above). `tsconfig.json` compiles `src/**/*` including tests, so the whole build fails; `main` builds clean. Bumping `@types/oracledb` doesn't help — 7.0.0 (latest) has the same gap. Note vitest's typecheck does **not** catch this; only `tsc`/`pnpm build` does, and this package has no GitHub CI.

Fix that compiles clean under the project tsconfig:

```ts
{ embedding: { val: queryVector, dir: oracledb.BIND_IN, type: oracledb.DB_TYPE_VECTOR } }
```

(or cast the binds object `as oracledb.BindParameters`).

libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts
```diff
@@ -1350,6 +1358,102 @@ describe("OracleVectorStore", () => {
     );
   });
 
+  test("uses vector index transform with JSON search index in the plan", async () => {
```
fede-kamel, line 1359: This test EXPLAINs a hand-written copy of the production SQL (hint included), so it verifies that the hint works in Oracle, but not that `OracleVS` actually emits it — if the hint were removed from `vectorstores.ts`, this test would still pass. The new unit test in `vectorstores.standard.test.ts` is what guards emission, which is fine, but worth a comment here noting that split — or better, generate the statement from the store so the two can't drift.

Also, the plan assertions (`IVF_FLAT_(CENTROIDS|CENTROID_PARTITIONS)`, the exact `TABLE ACCESS BY USER ROWID | testlangchain_1` spacing, and the hardcoded table name instead of `oraclevs.tableName`) are brittle across DB versions/plan formats. They do pass on 26ai today, but consider loosening them.

libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts
```diff
@@ -1245,6 +1245,14 @@ describe("OracleVectorStore", () => {
     oraclevs = new OracleVS(embedder, dbConfig);
     await oraclevs.initialize();
 
+    // create Index so that vector index is being used.
+    await createIndex(connection as oracledb.Connection, oraclevs, {
```
fede-kamel: Minor: here the IVF index is created *before* `addDocuments`, while the new EXPLAIN PLAN test below creates it *after* inserting. IVF centroids trained on an empty table plus `FETCH APPROX` introduces a small recall/nondeterminism risk in a test that asserts exact filter results (it passes today with 4 docs and k=4, but only because the approximate top-k covers the whole table). Creating the index after the inserts, as the other test does, would be consistent and safer. A one-line comment on why this particular filter test gets an index would also help.

libs/js/langchain-oracledb/src/tests/vectorstores.unit.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, expect, test, vi } from "vitest";
```
fede-kamel, line 1: Naming nit: the `*.standard.test.ts` pattern is the `standard-unit` bucket in `vitest.config.ts`, which by convention is for the LangChain standard-tests conformance suites. This is a bespoke SQL-generation unit test — a name like `vectorstores.unit.test.ts` (or just folding it into a plain `*.test.ts`) would keep that bucket clean. Cosmetic only; the tests themselves look good and pass.

fede-kamel (changes requested):
Thanks for the fix and for porting it from the Python side. I ran the full suite against an Oracle Autonomous Database 26ai instance (`lint`, `test`, `test:standard:unit`, `test:int`, `build`) and verified the hint behavior independently. Two issues need to be addressed before merge — see inline comments for details:

1. **`pnpm build` fails on this branch** (TS2769 in the new EXPLAIN PLAN test) — `main` builds clean, so this is introduced here. Since this package has no GitHub CI, nothing upstream catches it until release.
2. **The hint changes query *results*, not just the plan, when a metadata filter is present.** I reproduced a filtered similarity search returning 0 rows with the hint vs. 3 correct rows without it (post-filtering vs. pre-filtering). Details inline on `vectorstores.ts`.

Test results otherwise: unit 3/3 , integration 42 pass / 2 skip  (the new `DEMO_CREDENTIAL` gating works nicely), and the new EXPLAIN PLAN test passes.

One more thing, since this PR's goal is to let the suite run completely: `createIndex builds HNSW vector index with custom parameters` fails with `ORA-01745` on 26ai — `:table` and `:index` are Oracle reserved words used as bind variable names (`vectorstores.int.test.ts` ~line 182, pre-existing from #160, fails identically on `main`). Renaming those two binds (e.g. `:tab` / `:idx`) would be an easy drive-by fix here.

libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts
sudarshan12s: Thanks . 

I added a comment in the `EXPLAIN PLAN` test noting that it validates Oracle’s plan behavior for the same SQL shape, while the dedicated unit test now guards actual hint emission from `OracleVS`.

 I considered generating the statement from the store directly, but that would require reaching into query construction just for the test; for this PR I kept the smaller change and documented the split. Is this fine?


I also loosened the brittle plan assertions:
- removed the `IVF_FLAT_(CENTROIDS|CENTROID_PARTITIONS)` check
- removed the hardcoded `testlangchain_1`
- stopped depending on exact `TABLE ACCESS BY USER ROWID | ...` formatting

The test now checks for the vector index name, the presence of a `TABLE ACCESS BY USER ROWID` step, and the runtime table name in a case-insensitive way.

libs/js/langchain-oracledb/src/tests/vectorstores.unit.test.ts
sudarshan12s: Renamed as `vectorstores.unit.test.ts`

libs/js/langchain-oracledb/src/vectorstores.ts
sudarshan12s: I have changed to emit simple `select` when filter is present and with `hint` when no filter is given. 
 ```
 const selectClause = hasFilter
+        ? "SELECT"
+        : `SELECT /*+ VECTOR_INDEX_TRANSFORM(${this.tableName}) */`;
```

I also added to remove exception for `no rows` and instead returned empty array `[] `to keep inline with python..


One thing I came across while doing the changes . Should we also avoid `FETCH APPROX` for filtered searches or we should just document? 
```

const fetchClause = hasFilter
? `FETCH FIRST :${bindValues.length} ROWS ONLY`
: `FETCH APPROX FIRST :${bindValues.length} ROWS ONLY`;

```

File: libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts (hunk 2 above)
Comment (sudarshan12s) [reply libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts]: Thanks. I was only running `pnpm run test` which was working fine (` Tests  48 passed (48)) `
I fixed it taking the suggestion.

`{ embedding: { val: queryVector, dir: oracledb.BIND_IN, type: oracledb.DB_TYPE_VECTOR } }`

File: libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts (hunk 4 above)
Comment (sudarshan12s) [reply libs/js/langchain-oracledb/src/tests/vectorstores.int.test.ts]: I have moved `createIndex` after `addDocuments` and added a comment.

**Comment by sharadraju** [comment]
@fede-kamel Thank you for all the reviews. Can you please check and approve the PR?

File: libs/js/langchain-oracledb/src/vectorstores.ts (hunk 1 above)
Comment (fede-kamel) [reply libs/js/langchain-oracledb/src/vectorstores.ts]: Tested your `FETCH APPROX` question on a 26ai instance. Short answer: with the hint removed, **`FETCH APPROX` under a filter is benign — you don't need to switch it to fix this bug.**

Setup: 200 rows sitting on the query point with `category:"games"` (non-matching) + 200 far rows with `category:"books"`, IVF vector index + JSON search index, `k=5`, filter `category=="books"` (correct answer = 5 books):

| SQL shape | rows |
|---|---|
| no hint + `FETCH APPROX` + filter (this PR) | **5/5 books** ✅ |
| no hint + `FETCH FIRST` (exact) + filter | 5/5 books ✅ |

The `EXPLAIN PLAN` for the no-hint + `FETCH APPROX` path shows why it's safe — the optimizer pre-filters through the JSON domain index and the vector index isn't used at all, so the approximate fetch can't drop matching rows:

```
| 3 | SORT ORDER BY STOPKEY         |
| 4 |  TABLE ACCESS BY INDEX ROWID  | <table>
| 5 |   DOMAIN INDEX                | <json search index>
```

Exact `FETCH FIRST` (your other option) also returns 5/5 and has the appeal of being correct *by construction* — it doesn't rely on the optimizer choosing to pre-filter. I'd treat that as an optional hardening follow-up rather than part of this PR, and if you do it, mirror it in the Python port (#211) so the two stay in lockstep. For now a one-line comment documenting the approximate-under-filter semantics is plenty. Not blocking.

fede-kamel (approved):
Re-reviewed after `fe4d752` — all five threads are addressed, and I re-verified the two blockers from my earlier review, this time including a live run against an Autonomous Database 26ai instance.

**Build (the release blocker):** `pnpm build` now exits 0 — the `Float32Array` bound via `{ val, dir: BIND_IN, type: DB_TYPE_VECTOR }` typechecks under the project tsconfig. `pnpm lint` is clean. The renamed `vectorstores.unit.test.ts` is green and now guards **both** branches: hint present when unfiltered, absent when a filter is supplied.

**Hint correctness — bug reproduced and fix confirmed on 26ai.** Setup: 200 rows on the query point with `category:"games"` (non-matching) + 200 far rows with `category:"books"`, IVF vector index + JSON search index, `k=5`, filter `category=="books"` (correct answer = 5 books):

| SQL shape | rows | |
|---|---|---|
| hint + `FETCH APPROX` + filter (old `main`) | **0/5** | ❌ the reported bug |
| no hint + `FETCH APPROX` + filter (this PR) | **5/5 books** | ✅ |
| no hint + `FETCH FIRST` (exact) + filter | 5/5 books | ✅ |

Dropping the hint restores correct, exact results. The `EXPLAIN PLAN` for the new path shows the mechanism — with the hint gone the optimizer pre-filters through the JSON domain index (`DOMAIN INDEX` → `SORT ORDER BY STOPKEY`) rather than approximate-top-k-then-filter, so the vector index isn't on the filtered path at all.

Also confirmed: the `throw "No rows found."` → `[]` change matches the Python port, the `EXPLAIN PLAN` test assertions are appropriately loosened, the IVF index is now created after `addDocuments`, and the ORA-01745 reserved-word binds (`:table`/`:index`) are renamed.

LGTM — approving. One non-blocking note on the open `FETCH APPROX`-under-filter question is inline (it's benign with the hint removed; exact `FETCH FIRST` is an optional follow-up that should be mirrored in #211).

sudarshan12s:
> Re-reviewed after `fe4d752` — all five threads are addressed, and I re-verified the two blockers from my earlier review, this time including a live run against an Autonomous Database 26ai instance.
> 
> **Build (the release blocker):** `pnpm build` now exits 0 — the `Float32Array` bound via `{ val, dir: BIND_IN, type: DB_TYPE_VECTOR }` typechecks under the project tsconfig. `pnpm lint` is clean. The renamed `vectorstores.unit.test.ts` is green and now guards **both** branches: hint present when unfiltered, absent when a filter is supplied.
> 
> **Hint correctness — bug reproduced and fix confirmed on 26ai.** Setup: 200 rows on the query point with `category:"games"` (non-matching) + 200 far rows with `category:"books"`, IVF vector index + JSON search index, `k=5`, filter `category=="books"` (correct answer = 5 books):
> 
> SQL shape	rows	
> hint + `FETCH APPROX` + filter (old `main`)	**0/5**	❌ the reported bug
> no hint + `FETCH APPROX` + filter (this PR)	**5/5 books**	✅
> no hint + `FETCH FIRST` (exact) + filter	5/5 books	✅
> Dropping the hint restores correct, exact results. The `EXPLAIN PLAN` for the new path shows the mechanism — with the hint gone the optimizer pre-filters through the JSON domain index (`DOMAIN INDEX` → `SORT ORDER BY STOPKEY`) rather than approximate-top-k-then-filter, so the vector index isn't on the filtered path at all.
> 
> Also confirmed: the `throw "No rows found."` → `[]` change matches the Python port, the `EXPLAIN PLAN` test assertions are appropriately loosened, the IVF index is now created after `addDocuments`, and the ORA-01745 reserved-word binds (`:table`/`:index`) are renamed.
> 
> LGTM — approving. One non-blocking note on the open `FETCH APPROX`-under-filter question is inline (it's benign with the hint removed; exact `FETCH FIRST` is an optional follow-up that should be mirrored in #211).

Thanks so much @fede-kamel for your time for checking in detail.

PR #237, opened 2026-06-17
feat(oci): surface streaming reasoning as standard content blocks

### LINKED ISSUE #238: ChatOCIGenAI: streaming responses drop model reasoning (no `reasoning` content block)
### Problem
`ChatOCIGenAI` exposes model reasoning as a langchain-core v1 `reasoning` content block on the **non-streaming** path (`invoke`): `reasoning_content` is placed in `additional_kwargs`, and `AIMessage.content_blocks` lifts it into a `reasoning` block.

On the **streaming** path it is dropped — `chat_stream_generation_info` returns only `finish_reason`, and per-delta reasoning is never captured. So a streamed message's `.content_blocks` has no `reasoning` block, unlike `invoke()`.

### Impact
Inconsistent behavior between `invoke` and `stream`; downstream code relying on standard content blocks (reasoning UIs, agent middleware) loses reasoning when streaming.

### Expected
Streaming should surface reasoning at parity with `invoke()` — a `reasoning` content block in the merged streamed message — without changing `.content` (backward compatible).

fede-kamel:
## Summary

`ChatOCIGenAI` already exposed model reasoning as a langchain-core v1 `reasoning`
content block on the **non-streaming** path (the `reasoning_content` rides in
`additional_kwargs`, which `AIMessage.content_blocks` lifts into a `reasoning`
block). The **streaming** path dropped it entirely — `chat_stream_generation_info`
returned only `finish_reason`, so a streamed message's `.content_blocks` had no
reasoning block even though `invoke()` did.

This PR brings streaming to parity.

## What changed

- **`Provider.chat_stream_to_reasoning`** — new hook with a base **no-op default**
  (`""`), so providers that don't override it (Cohere, data-science) are unchanged.
- **`GenericProvider`** overrides it to read the per-delta `message.reasoningContent`
  field. OCI streaming events are raw JSON with **camelCase** keys
  (`finishReason` / `toolCalls` / `reasoningContent`), unlike the **snake_case**
  non-streaming SDK object (`reasoning_content`) — verified live.
- **`_stream` / `_astream`** accumulate the delta into
  `AIMessageChunk.additional_kwargs["reasoning_content"]` (only when non-empty),
  which concatenates across chunks, so the merged message exposes a standard
  `reasoning` content block.

## Backward compatibility

Fully backward compatible:
- `.content` stays a `str` — no `output_version="v1"` / `model_provider` change.
- The `reasoning_content` key appears only when a model actually emits reasoning.
- Other providers are byte-identical via the base default.
- On langchain-core < 1.0 (py3.9 matrix) the key just sits harmlessly in
  `additional_kwargs`; the content-block tests skip.

## Testing

**Unit** (`tests/unit_tests/chat_models/`): added a content-blocks contract suite
(text / reasoning sync+stream / tool_call, plus an explicit assertion that
`.content` stays a `str`) and provider-level reasoning-extraction tests.
Full unit suite: **430 passed, 12 skipped**.

**Integration** (`tests/integration_tests/chat_models/test_chat_features.py`,
run live against OCI Generative AI, us-chicago-1): added live reasoning
content-block + streaming-reasoning assertions. Entire file: **28 passed**
across `xai.grok-3-mini-fast`, `openai.gpt-oss-120b`, `meta.llama-3.3-70b-instruct`,
`cohere.command-r-08-2024` — including all pre-existing chat features (chains,
stream/astream, tool calling, structured output), confirming no regression.

## Note: gpt-oss-120b reclassification

`conftest.py` moves `openai.gpt-oss-120b` from the standard list to the reasoning
list. The OCI deployment now surfaces reasoning on **both** `invoke` and `stream`;
it was previously null on `invoke`, which is why it had been listed as standard.
This is a test-fixture correction only — no library behavior change.

Closes #238

### DIFF
--- a/libs/oci/langchain_oci/chat_models/async_mixin.py
+++ b/libs/oci/langchain_oci/chat_models/async_mixin.py
@@ -201,9 +201,19 @@ class ChatOCIGenAIAsyncMixin:
                     event_data, tool_call_ids
                 )
 
+                # Surface incremental reasoning (e.g. xAI Grok, OpenAI o-series)
+                # so the merged message exposes a standard ``reasoning`` content
+                # block, matching the non-streaming path. Absent for models that
+                # don't emit reasoning, in which case nothing is added.
+                reasoning_delta = self._provider.chat_stream_to_reasoning(event_data)  # type: ignore[attr-defined]
+                additional_kwargs = (
+                    {"reasoning_content": reasoning_delta} if reasoning_delta else {}
+                )
+
                 chunk = ChatGenerationChunk(
                     message=AIMessageChunk(
                         content=delta,
+                        additional_kwargs=additional_kwargs,
                         tool_call_chunks=tool_call_chunks,
                     )
                 )
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -564,9 +564,19 @@ class ChatOCIGenAI(ChatOCIGenAIAsyncMixin, BaseChatModel, OCIGenAIBase):
                     event_data, tool_call_ids
                 )
 
+                # Surface incremental reasoning (e.g. xAI Grok, OpenAI o-series)
+                # so the merged message exposes a standard ``reasoning`` content
+                # block, matching the non-streaming path. Absent for models that
+                # don't emit reasoning, in which case nothing is added.
+                reasoning_delta = self._provider.chat_stream_to_reasoning(event_data)
+                additional_kwargs = (
+                    {"reasoning_content": reasoning_delta} if reasoning_delta else {}
+                )
+
                 chunk = ChatGenerationChunk(
                     message=AIMessageChunk(
                         content=delta,
+                        additional_kwargs=additional_kwargs,
                         tool_call_chunks=tool_call_chunks,
                     )
                 )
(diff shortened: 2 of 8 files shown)


fede-kamel:
### Additional live validation (probes)

Beyond the unit + integration tests in the PR, I ran targeted live probes against OCI Generative AI (`xai.grok-3-mini-fast`, us-chicago-1) on the paths most affected by this change:

| Probe | Focus | Result |
|---|---|---|
| Sync stream | reasoning surfaces, `.content` stays `str`, block order | blocks `['reasoning','text']` ✅ |
| **Async stream (`_astream`)** | the async streaming path end-to-end | reasoning surfaced, blocks `['reasoning','text']` ✅ |
| **Reasoning + tool call (streaming)** | the per-delta `additional_kwargs["reasoning_content"]` shares the chunk with `tool_call_chunks` — verify it doesn't break tool accumulation | `tool_calls=[get_weather(city='Paris')]` intact, blocks `['reasoning','tool_call']` ✅ |
| invoke vs stream parity | same prompt → same block shape | both `['reasoning','text']` ✅ |

Key points confirmed: `_astream` works end-to-end (not just the mirrored unit test); reasoning and tool calls coexist correctly in the merged message; the `reasoning` block is consistently ordered first; and streaming matches `invoke()`.

fileames:
Could we add a small async regression for this path as well? `_astream` now has the same `reasoning_content` accumulation logic as `_stream`, but the new mocked regression only exercises sync streaming. A mocked `chat_async` event with `message.reasoningContent`, merged through `_astream`, would make sure async streaming stays at parity too.

fileames (approved):

PR #232, opened 2026-06-02
[chore] Bump up langchain oci to 0.2.7

YouNeedCryDear:

PR #246, opened 2026-06-24
bump up langchain oci to 0.2.8

fede-kamel:
Bumps the `langchain-oci` package version `0.2.7` → `0.2.8` so the recently merged work can be cut as a release (e.g. #237 — streaming reasoning surfaced as standard content blocks).

Version-only change to `libs/oci/pyproject.toml`, following the cadence of #232 / #222.

### DIFF
--- a/libs/oci/pyproject.toml
+++ b/libs/oci/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "langchain-oci"
-version = "0.2.7"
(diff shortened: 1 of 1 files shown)


fede-kamel:
Superseded: main is already at 0.3.0 (#251), and #254 carries the 0.3.1 bump for the next bugfix release, so the 0.2.8 bump is obsolete.

PR #250, opened 2026-06-30
feat(js/langchain-oracledb): Add oracledb persistence for Langgraph.js

Devx228:
## Summary
adds @oracle/langgraph-oracledb, a new TypeScript package for persisting LangGraph.js applications in Oracle Database.

It provides:
  -OracleCheckpointSaver for saving and resuming LangGraph checkpoints, pending writes, and channel state.
  -OracleStore for persistent key-value storage, metadata filtering, namespace operations, and vector similarity search.
  -Oracle schema migrations with idempotent and concurrency-safe setup.
  -Read-only diagnostics for inspecting schema, migrations, runtime configuration, and VECTOR support.
  -Native Oracle VECTOR support with compatibility fallbacks.
  -Support for caller-provided connections, pools, and configurable table prefixes.
  -Unit, integration, parity, and shared LangGraph checkpoint-validation coverage.

The package has been validated with 39 unit tests and 793 Oracle-backed integration tests, along with TypeScript, ESLint, formatting, dependency-cycle, and package-build checks.


libs/js/langgraph-oracledb/src/saver.ts
```diff
@@ -0,0 +1,1251 @@
...
+        this.sql.UPSERT_CHECKPOINTS_SQL,
+        {
+          thread_id: threadId,
+          checkpoint_ns: encodedCheckpointNs,
+          checkpoint_id: checkpoint.id,
+          parent_checkpoint_id: parentCheckpointId,
+          type: checkpointType,
+          metadata_type: metadataType,
+          checkpoint: useClobStorage
+            ? new TextDecoder().decode(checkpointBytes)
```
fede-kamel, line 681: **Blocker (reproduced e2e): non-fatal `TextDecoder` silently destroys binary serde payloads in CLOB mode.**

Repro against Oracle 23ai Free: created the checkpoint schema with `checkpoint`/`metadata` as `CLOB` (the pre-existing-schema case this fallback exists for), then used a gzip serde — legal per `SerializerProtocol`, same shape as any msgpack/compressed serializer:

- `put()` succeeds with **no error** — invalid UTF-8 sequences are replaced with U+FFFD during `decode()`.
- `getTuple()` then throws `Z_DATA_ERROR: incorrect header check` from `loadCheckpoint` — the checkpoint is permanently unrecoverable.

The README's "text-compatible serializer payloads only" caveat isn't enforced anywhere at runtime. Suggest `new TextDecoder("utf-8", { fatal: true })` so the write *fails loudly* instead of corrupting, or refusing CLOB mode outright when the serde type isn't a known text format.

libs/js/langgraph-oracledb/src/sql.ts
```diff
@@ -0,0 +1,432 @@
...
+  if (!checkpointNs.startsWith("b64:")) {
+    return checkpointNs;
+  }
+  return Buffer.from(checkpointNs.slice(4), "base64url").toString("utf8");
+};
+
+export const validateTablePrefix = (
+  tablePrefix: string = DEFAULT_TABLE_PREFIX
+): string => {
+  if (tablePrefix === "") return tablePrefix;
```
fede-kamel, line 77: **Blocker (reproduced e2e): `tablePrefix: ""` is allowed and collides with the Python `langgraph-oracledb` schema — which is not wire-compatible.**

Empty prefix yields `CHECKPOINTS` / `CHECKPOINT_BLOBS` / `CHECKPOINT_WRITES` / `CHECKPOINT_MIGRATIONS` — exactly the table names the Python package creates. But Python stores `checkpoint`/`metadata` as native `JSON` columns (no `metadata_type`, different ns sentinel, extra `task_path`), while this package expects BLOBs + `metadata_type`.

Repro against Oracle 23ai Free:
1. Python `OracleSaver(conn).setup()` + one `put()` → tables created, migration versions 0–6 recorded.
2. JS `new OracleCheckpointSaver({ connection, tablePrefix: "" }); await saver.setup()` → **succeeds silently**: Python's recorded migration version (6) exceeds this package's highest (5), so every migration is skipped and the schema is assumed current.
3. First `saver.put(...)` → `ORA-00904: "METADATA_TYPE": invalid identifier` at runtime.

So a user pointing both SDKs at one schema gets a working `setup()` and a broken runtime. Suggest (a) rejecting the empty prefix, (b) having `setup()`/`detectCheckpointStorageMode` fail fast when `CHECKPOINTS.checkpoint` is a `JSON` column (that's unambiguously the Python schema), and (c) a README note that the JS and Python packages are not storage-compatible.

libs/js/langgraph-oracledb/src/oracledb.d.ts
```diff
@@ -0,0 +1,72 @@
+declare module "oracledb" {
```
fede-kamel, line 1: node-oracledb 6/7 ships its own TypeScript declarations and `@types/oracledb` is already in devDependencies — a hand-rolled ambient `declare module "oracledb"` shadows those, risks drifting from the real API (the `execute`/`Pool` signatures here are much narrower than reality), and can break downstream compilation depending on declaration merge order once the package publishes its `dist` types. Prefer the upstream types; if a narrowed surface is wanted, local interfaces like `OracleConnectionLike` in `saver.ts` already do that without the ambient override.

fede-kamel (changes requested):
This is a strong, carefully engineered first drop — bind variables and identifier validation throughout, real transactions with rollback, the shared `@langchain/langgraph-checkpoint-validation` suite wired in, and thoughtful Oracle-specific handling (empty-ns sentinels, ORA-error idempotency, MERGE upserts, LOB fetchInfo). Thank you!

I reviewed the full diff against the Python `langgraph-oracledb` reference and **reproduced the two data-level findings end-to-end against a real Oracle 23ai Free database** — repro details in the inline comments. Requesting changes on:

1. **Rebase needed** — `README.md` conflicts with `main` (GitHub reports the PR as CONFLICTING).
2. **CLOB fallback silently corrupts binary serde payloads** (reproduced live — see inline comment in `saver.ts`).
3. **Empty `tablePrefix` collides with the Python package's schema**, which is not wire-compatible; `setup()` accepts it silently and the first write fails at runtime (reproduced live — see inline comment in `sql.ts`).
4. **Hand-rolled `src/oracledb.d.ts` ambient module** — see inline comment.

Smaller items, fine as follow-ups but worth tracking:
- Lazy pool creation race: two concurrent first calls to `getConnection()`/`ensurePool()` both see no pool and create two; one leaks past `end()`. Guard with a cached promise like `setupPromise`.
- `list()` fetches all rows into memory, then issues 2–3 queries per checkpoint (each acquiring its own pooled connection); with a metadata `filter` the SQL `FETCH FIRST` is dropped entirely and filtering happens client-side. The Python impl pushes JSON predicates into SQL.
- Channel blob upserts run one `execute` per row inside `put()` while the store side already uses `executeMany` — same batching would help here.
- Key columns are `VARCHAR2(512)` with a 512-byte validation; Python uses `VARCHAR2(2000)` and Postgres uses TEXT, so thread IDs valid elsewhere throw here (and the base64url ns-encoding inflates sizes ~33%). Consider 2000 for parity or document the limit.
- `resolveVectorBindStrategy` latches `vectorBindStrategy = "string"` permanently on *any* probe error, including transient connection failures — consider latching only on the specific unsupported-type error.
- No CI runs any of this package's tests (the repo workflows only cover the Python libs). At minimum the unit tests (`pnpm test`, no DB needed) should be wired into CI.
- `LICENSE` says "Copyright (c) 2023 LangChain" (MIT) on new Oracle-authored code in a UPL-1.0 repo — only the `scripts/*.js` build helpers look derived from langchainjs. Flagging for maintainer confirmation.
- The nested `pnpm-workspace.yaml` has no `packages` field, so a plain `pnpm install` inside the package fails with "packages field missing or empty" (hit this while setting up the repro).
- `TASKS_CHANNEL = "__pregel_tasks"` in `sql.ts` duplicates the `TASKS` constant already imported from `@langchain/langgraph-checkpoint` in `saver.ts`.

fede-kamel:
@Devx228 two follow-ups on the review:

**1. E2E verification.** I reproduced both data-level findings against a live Oracle 23ai Free database (gvenzl/oracle-free:23-slim, node-oracledb 7 thin), on your branch both before and after rebasing on main:
- **CLOB corruption**: with a `CHECKPOINTS` table whose `checkpoint`/`metadata` are CLOB and a gzip serde, `put()` succeeds silently and `getTuple()` throws `Z_DATA_ERROR: incorrect header check` — the checkpoint is unrecoverable.
- **Python schema collision**: after Python `langgraph-oracledb` `OracleSaver.setup()` creates its unprefixed schema (migration versions 0–6), JS `new OracleCheckpointSaver({ tablePrefix: "" }).setup()` succeeds silently (6 > this package's max version 5, so all migrations are skipped), then the first `put()` fails with `ORA-00904: "METADATA_TYPE": invalid identifier`.

**2. Rebase done for you.** Since the PR conflicts with main (README.md), I rebased your 8 commits onto current `main` and resolved the conflict by keeping main's restructured README and adding an `@oracle/langgraph-oracledb` row to the Packages table. The result is at [`fede-kamel:pr250-rebased-on-main`](https://github.com/fede-kamel/langchain-oracle/tree/pr250-rebased-on-main) — both repros behave identically on it and your code is otherwise untouched. If it looks good to you, you can update this PR with:

```bash
git remote add fede https://github.com/fede-kamel/langchain-oracle.git
git fetch fede pr250-rebased-on-main
git checkout add-js-langgraph-oracledb
git reset --hard fede/pr250-rebased-on-main
git push --force-with-lease origin add-js-langgraph-oracledb
```

One more small thing I hit while setting up the repro: `pnpm install` inside `libs/js/langgraph-oracledb` fails with "packages field missing or empty" because of the nested `pnpm-workspace.yaml` — worth fixing alongside the review items.

Devx228:
Hiii @fede-kamel , thank you for the exceptionally detailed review , reproducing both findings live against Oracle 23ai free db and rebasing my branch onto main is far more than I expected on a draft PR, and I'm grateful for it.

I want to be upfront: my internship at Oracle has ended, so I am doubtful if I can continue this PR myself. Elif (@fileames) is now working in the main repo and could pick this up or I myself can work on it.

Before we decide how to proceed, I wanted to ask what you'd prefer. A few options I can think of: Elif takes over this branch directly and addresses your review items; or I can continue with my branch resolving these issues,  or you have another approach in mind. Whatever's cleanest for you and the repo works for me.

Let me know and I'll help coordinate so the review context and repro details carry over to Elif. Thanks again for all the effort you have put for this. I would be extremely happy if I myself could work on it.

fileames:
Hi @Devx228 and @fede-kamel, we are working on merging the code to the main LangGraph first: https://github.com/langchain-ai/langgraphjs/pull/2660 
I will apply the changes needed for this review there.
Thanks for all the help :)

fileames:
@fede-kamel Made the changes requested to the PR on the original langgraphjs repo: https://github.com/langchain-ai/langgraphjs/pull/2660. Would be happy to address anything remaining :)

fede-kamel:
Closing in favour of the upstream submission, langchain-ai/langgraphjs#2660, where @fileames is carrying the review feedback from here. Thank you @Devx228 for the original work; the Oracle checkpointer for LangGraph.js will land upstream and this repo will point users there. If upstream stalls we can reopen.

Devx228:
Yess Thankyou @fede-kamel for the review and absolutely grateful to @fileames for her support and efforts into this. Hope our package lands there very soon! ;)

PR #252, opened 2026-07-02
Fix incorrect tool-call indices during streaming

joseph-klein:
## Summary

Fix GPT streamed parallel tool-call reconstruction in `langchain-oci`.

GPT models can stream tool calls as:

1. A start chunk with `id`, `name`, and empty `arguments`
2. Many following argument chunks that contain only `arguments` fragments, without `id` or `name`

`GenericProvider.process_stream_tool_calls` was using the per-event `enumerate` index as the stable tool-call identity. That breaks when GPT emits multiple tool calls in one assistant turn, because many argument-fragment events contain only one `toolCalls[0]` item.

## Problem

With one streamed tool call, id-less argument fragments can usually be mapped back to the only open tool call.

With parallel tool calls, GPT can open tool A, stream A's argument fragments, then open tool B and stream B's argument fragments. Those B fragments are also id-less and also appear as `toolCalls[0]` in single-item stream events.

Example:

```text
event 0:  toolCalls[0] = {id: call_A, name: market_research_tool, arguments: ""}
event 1:  toolCalls[0] = {arguments: "{\"idea"}
...
event 21: toolCalls[0] = {id: call_B, name: customer_signal_tool, arguments: ""}
event 22: toolCalls[0] = {arguments: "{\"idea"}
```

The old logic could assign event 22 to `call_A` instead of `call_B`, leaving later tools with empty `{}` args and causing required-parameter validation failures.

## Fix

Track the active stream tool-call index when a chunk with an `id` opens a new tool call. When a later single id-less argument chunk arrives, attach it to the active tool call.

## Validation

Validated with a local GPT-4.1 repro using three parallel tools.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -722,13 +722,27 @@ class GenericProvider(Provider):
         if not tool_call_response:
             return tool_call_chunks
 
-        for idx, tool_call in enumerate(
-            self.format_stream_tool_calls(tool_call_response)
-        ):
+        formatted_tool_calls = self.format_stream_tool_calls(tool_call_response)
+        single_idless_chunk = (
+            len(formatted_tool_calls) == 1 and not formatted_tool_calls[0].get("id")
+        )
+        active_index = getattr(self, "_active_stream_tool_call_index", None)
+
+        for idx, tool_call in enumerate(formatted_tool_calls):
             tool_id = tool_call.get("id")
 
             if tool_id:
-                if idx not in tool_call_ids or tool_call_ids[idx] == tool_id:
+                existing_index = next(
+                    (
+                        stored_index
+                        for stored_index, stored_id in tool_call_ids.items()
+                        if stored_id == tool_id
+                    ),
+                    None,
+                )
+                if existing_index is not None:
+                    index = existing_index
+                elif idx not in tool_call_ids or tool_call_ids[idx] == tool_id:
                     # New idx - use idx as index
                     # Same ID at same idx - reuse idx as index
                     index = idx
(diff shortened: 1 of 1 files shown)


adoerfle:
https://github.com/oracle/langchain-oracle/issues/253

fede-kamel:
Thanks for this — you diagnosed the root cause first, and your analysis was exactly right. Closing in favor of #254, which lands the same active-tool-call routing approach with a per-position map wired into the per-stream reset lifecycle, plus a unit regression test replaying the #253 stream shape and a live integration test verified against openai.gpt-4.1 (it fails on pre-fix main with the exact reported symptom). It also carries the 0.3.1 release bump so the fix ships to PyPI promptly.

One idea from this PR worth keeping: the reverse id→index lookup for re-sent tool-call ids at a different array position — good hardening, happy to fold it into a follow-up with a test.

A review on #254 would be very welcome if you have cycles. (Also flagging: the OCA check here was failing — worth signing the Oracle Contributor Agreement for future contributions.)

fede-kamel:
Follow-up with verification results, for the record — I probed this patch directly rather than just asserting the overlap:

- **Correctness: confirmed.** I checked out this PR's head and ran the #254 regression suite against it: the three-tool GPT parallel stream shape from #253 reconstructs correctly (4/4 applicable unit tests pass), and the live integration test against `openai.gpt-4.1` passes as well. Your fix works.
- **What blocked it instead:** the branch would fail repo lint (`ruff format` flags `generic.py` — missing EOF newline), it carries no tests of its own, it's based on a ~March main, and the OCA check is unsigned, which prevents merging regardless.

So to be explicit: this was closed for redundancy with the already-green, tested, release-carrying #254 — not because the approach was wrong. It wasn't; you had the right diagnosis and a working fix first.

PR #254, opened 2026-07-07
fix(oci): attach id-less streamed tool-call argument fragments to the active tool call

### LINKED ISSUE #253: GPT streamed parallel tool-call argument fragments can be assigned to the wrong tool call
## Summary

GPT models can stream parallel tool calls as:

1. A start chunk with `id`, `name`, and empty `arguments`
2. Following chunks that contain only `arguments` fragments, without `id` or `name`

`GenericProvider.process_stream_tool_calls` appears to use the per-event `enumerate` index as the stable tool-call identity. That breaks when GPT emits multiple tool calls in one assistant turn, because many argument-fragment events contain only one `toolCalls[0]` item.

## Problem

Example stream shape:

```text
event 0:  toolCalls[0] = {id: call_A, name: market_research_tool, arguments: ""}
event 1:  toolCalls[0] = {arguments: "{\"idea"}
...
event 21: toolCalls[0] = {id: call_B, name: customer_signal_tool, arguments: ""}
event 22: toolCalls[0] = {arguments: "{\"idea"}
```

The current logic can assign event 22 to `call_A` instead of `call_B`, leaving later tools with empty `{}` args or malformed reconstructed arguments. Tools with required params then fail validation even though the model produced the arguments.

## Repro

Use a GPT model with streaming enabled and parallel tool calls enabled. Bind multiple tools with required parameters, then prompt the model to call more than one tool in the same assistant turn.

Example:

```text
Call all three tools before responding:
- market_research_tool with idea and marketSegment
- customer_signal_tool with idea and audience
- roadmap_risk_tool with productArea and riskFocus
```

Observed result: one or more later tool calls may be reconstructed with `{}` args or wrong args.

## Expected Behavior

Id-less argument fragments should be associated with the logical tool call that was most recently opened by a chunk containing `id` and `name`.

## Suggested Fix

Track the active stream tool-call index when a chunk with an `id` opens a new tool call. When a later single id-less argument chunk arrives, attach it to the active tool call instead of using the per-event `enumerate` index.

fede-kamel:
## Summary

Fixes #253 — GPT streamed parallel tool-call argument fragments can be assigned to the wrong tool call.

GPT models served via OCI GenAI stream parallel tool calls **sequentially at `toolCalls[0]`**: each call is opened by a chunk carrying `id`/`name` and empty `arguments`, followed by argument fragments that carry neither `id` nor `name`:

```text
event 0:  toolCalls[0] = {id: call_A, name: market_research_tool, arguments: ""}
event 1:  toolCalls[0] = {arguments: "{\"idea"}
...
event 21: toolCalls[0] = {id: call_B, name: customer_signal_tool, arguments: ""}
event 22: toolCalls[0] = {arguments: "{\"idea"}      <-- belongs to call_B
```

## Root cause

`GenericProvider.process_stream_tool_calls` routed id-less fragments by their per-event array position (`enumerate` index):

- Event 21 (`call_B` at position 0, different id than `call_A`) was correctly detected as a parallel call and assigned logical index 1.
- But event 22's id-less fragment — still arriving at position 0 — matched the `elif idx in tool_call_ids` branch and was routed back to **logical index 0 (`call_A`)**.

Result: the first tool call accumulated both tools' argument fragments concatenated (malformed JSON), and later tool calls finished with empty `{}` args, failing tool validation downstream even though the model produced complete arguments.

The OCI stream payload carries no `index` field on tool calls (the `FunctionCall` model has only `id`, `name`, `arguments`, `type`), so array position alone cannot identify the logical call.

## Fix

Track, per stream, which logical tool-call index each array position currently feeds (`_active_tool_call_indices: Dict[int, int]` on the provider):

- When a chunk with an `id` opens (or re-opens) a call at position `idx`, record `position → logical index`.
- Id-less argument fragments at `idx` route through that map — i.e. they attach to the call most recently opened at that position — instead of assuming position equals identity.

The map is cleared by `reset_stream_state()`, the same per-stream lifecycle as the existing `_xml_buffer`, so both `_stream` and `_astream` get the fix and the reset automatically.

### Why this works — state trace over the #253 stream

Walking the fixed algorithm over the reported stream, with the two pieces of state: `tool_call_ids` (logical index → id, unchanged) and the new `_active_tool_call_indices` (array position → logical index):

| Event | Chunk at `toolCalls[0]` | Branch taken | position map | `tool_call_ids` | Emitted `(id, index)` |
|---|---|---|---|---|---|
| 0 | `id=call_A, name, args=""` | new id at new position → index 0 | `{0: 0}` | `{0: call_A}` | `(call_A, 0)` ✅ |
| 1..n | `args` fragment (no id) | position 0 → active index **0** | `{0: 0}` | `{0: call_A}` | `(call_A, 0)` ✅ |
| 21 | `id=call_B, name, args=""` | different id at used position → parallel call, index `len(tool_call_ids)` = 1 | `{0: 1}` | `{0: call_A, 1: call_B}` | `(call_B, 1)` ✅ |
| 22.. | `args` fragment (no id) | position 0 → active index **1** | `{0: 1}` | `{0: call_A, 1: call_B}` | `(call_B, 1)` ✅ **(was `(call_A, 0)` ❌ before)** |

The single divergence from the old logic is the fragment-routing step (event 22+): the old code looked up `tool_call_ids` by array position and always resolved position 0 to `call_A`; the new code looks up the position map, which event 21 repointed at `call_B`. Everything downstream is untouched — LangChain merges `ToolCallChunk`s by `index`, so once every fragment carries the right logical index, each tool call's arguments concatenate into complete, parseable JSON.

Correctness of the invariant: a fragment with no `id` can only belong to a call that was previously opened by an id-bearing chunk **at the same array position** — the API gives no other association. The map records exactly that (most recent id-bearing open per position), so fragments follow their opener by construction. For streams where positions never get reused (gpt-oss, Gemini), the map is the identity mapping and behavior is bit-for-bit identical to before.

### Compatibility with existing stream shapes

| Provider pattern | Stream shape | Behavior |
|---|---|---|
| GPT parallel (#253) | sequential opens at position 0, id-less fragments between | **fixed** — fragments follow the active call |
| gpt-oss fragmented | single call, id in first chunk only | unchanged — position maps to itself |
| Grok parallel | complete call per event, same position, different ids | unchanged — new logical index per new id |
| Gemini | id-less complete calls | unchanged — generated UUID per position |

## Proof

Two independent lines of evidence, one synthetic and one against the live service:

1. **Unit-level**: the new regression test replays the exact #253 stream shape (three calls opened sequentially at position 0 with id-less fragments between). Against the old code it fails — `call_A` receives all three tools' fragments and `call_B`/`call_C` merge with empty args; against this fix every chunk carries the right `(id, index)` and the merged message has three complete tool calls.
2. **Live A/B**: the new integration test was run against OCI GenAI with `openai.gpt-4.1` twice — once with `generic.py` from `main` (pre-fix) and once with this branch. Pre-fix it fails with the exact reported symptom; with the fix it passes on both sync and async paths:

```text
# generic.py from main (pre-fix):
AssertionError: openai.gpt-4.1: customer_signal_tool reconstructed with
empty args (fragments were routed to another tool call)
1 failed in 5.29s

# this branch:
2 passed in 5.00s
```

### Unit

- New regression test reproducing the exact #253 stream shape with three parallel tool calls (`call_A`/`call_B`/`call_C` opened sequentially at position 0 with id-less fragments between): asserts per-chunk `(id, index)` routing and the final LangChain-merged message — three tool calls, each with its own complete, parseable arguments.
- New test that `reset_stream_state()` clears the position map between streams.
- Full `tests/unit_tests` suite: **456 passed, 12 skipped**. ruff + mypy clean on 3.9 (langchain-core 0.3.x) and 3.10+ (langchain-core 1.x).

### Integration (live against OCI GenAI)

Added `tests/integration_tests/chat_models/test_openai_tool_calling.py`:

- Binds three `StructuredTool`s with required params (`market_research_tool`, `customer_signal_tool`, `roadmap_risk_tool` — mirroring the #253 repro) to a GPT model.
- Streams a single assistant turn that calls all three, merges chunks the way LangChain does, and asserts: no `invalid_tool_calls`, every tool call has non-empty args that validate against its tool's schema, and all tool-call ids are unique.
- Covers **both the sync (`stream`) and async (`astream`) paths**.
- Fully env-driven (`OCI_COMPARTMENT_ID`, `OCI_REGION`, `OCI_AUTH_TYPE`, `OCI_CONFIG_PROFILE`, `OCI_GPT_PARALLEL_MODELS` — default `openai.gpt-4.1`); skips when no compartment is configured.

Live results with `openai.gpt-4.1` are in the **Proof** section above: pre-fix code fails with the exact #253 symptom; this branch passes on both paths.

## Release

Also bumps `langchain-oci` to **0.3.1** for this bugfix release (0.3.0 is the current PyPI release).

### DIFF
--- a/libs/oci/langchain_oci/chat_models/async_mixin.py
+++ b/libs/oci/langchain_oci/chat_models/async_mixin.py
@@ -200,6 +200,9 @@ class ChatOCIGenAIAsyncMixin:
             messages, stop, stream=True, **kwargs
         )
         tool_call_ids: Dict[int, str] = {}
+        # Per-stream position -> logical index routing map, caller-owned so
+        # concurrent streams can't corrupt each other (see _stream's note).
+        active_tool_call_indices: Dict[int, int] = {}
 
         # Per-stream provider state, owned by this call (see _stream's note).
         # Concurrent astream calls on a shared chat-model instance interleave
@@ -256,7 +259,10 @@ class ChatOCIGenAIAsyncMixin:
                     event_data, **state_kwargs
                 )
                 tool_call_chunks = self._provider.process_stream_tool_calls(  # type: ignore[attr-defined]
-                    event_data, tool_call_ids, **state_kwargs
+                    event_data,
+                    tool_call_ids,
+                    active_tool_call_indices=active_tool_call_indices,
+                    **state_kwargs,
                 )
 
                 # Surface incremental reasoning (e.g. xAI Grok, OpenAI o-series)
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -589,6 +589,10 @@ class ChatOCIGenAI(ChatOCIGenAIAsyncMixin, BaseChatModel, OCIGenAIBase):
         request = self._prepare_request(messages, stop=stop, stream=True, **kwargs)
         response = self._chat_with_param_retry(request)
         tool_call_ids: Dict[int, str] = {}
+        # Per-stream toolCalls position -> logical index routing map, owned
+        # here (not on the shared provider) so concurrent streams on one
+        # ChatOCIGenAI instance can't corrupt each other's tool-call routing.
+        active_tool_call_indices: Dict[int, int] = {}
 
         # Per-stream provider state (currently the GenericProvider's
         # incremental <tool_call> XML buffer used for Hermes/Llama-style DAC
@@ -613,7 +617,10 @@ class ChatOCIGenAI(ChatOCIGenAIAsyncMixin, BaseChatModel, OCIGenAIBase):
                 # Process streaming content
                 delta = self._provider.chat_stream_to_text(event_data, **state_kwargs)
                 tool_call_chunks = self._provider.process_stream_tool_calls(
-                    event_data, tool_call_ids, **state_kwargs
+                    event_data,
+                    tool_call_ids,
+                    active_tool_call_indices=active_tool_call_indices,
+                    **state_kwargs,
                 )
 
                 # Surface incremental reasoning (e.g. xAI Grok, OpenAI o-series)
--- a/libs/oci/langchain_oci/chat_models/providers/base.py
+++ b/libs/oci/langchain_oci/chat_models/providers/base.py
@@ -158,11 +158,17 @@ class Provider(ABC):
         event_data: Dict,
         tool_call_ids: Dict[int, str],
         stream_state: Optional[Any] = None,
+        active_tool_call_indices: Optional[Dict[int, int]] = None,
     ) -> List[ToolCallChunk]:
         """Process streaming tool calls from event data into chunks.
 
         ``stream_state`` is the object returned by :meth:`new_stream_state`
         for the current stream (``None`` for stateless providers).
+
+        ``tool_call_ids`` and ``active_tool_call_indices`` are per-stream
+        aggregation state owned by the caller (``_stream`` / ``_astream``),
+        never by the provider, so concurrent streams sharing one provider
+        instance can't corrupt each other's tool-call routing.
         """
         ...
 
--- a/libs/oci/langchain_oci/chat_models/providers/cohere.py
+++ b/libs/oci/langchain_oci/chat_models/providers/cohere.py
@@ -739,6 +739,7 @@ class CohereProvider(Provider):
         event_data: Dict,
         tool_call_ids: Dict[int, str],
         stream_state: Optional[Any] = None,
+        active_tool_call_indices: Optional[Dict[int, int]] = None,
     ) -> List[ToolCallChunk]:
         """
         Process Cohere stream tool calls and return them as ToolCallChunk objects.
@@ -747,6 +748,9 @@ class CohereProvider(Provider):
             event_data: The event data from the stream
             tool_call_ids: Dict mapping tool call IDs for aggregation
             stream_state: Unused — Cohere parsing is stateless per event
+            active_tool_call_indices: Accepted for signature parity with the
+                base provider; Cohere streams don't emit id-less fragments at
+                reused positions, so it is unused here.
 
         Returns:
             List of ToolCallChunk objects
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -141,6 +141,13 @@ class GenericProvider(Provider):
         # and never touch it.
         self._xml_buffer = XmlStreamBuffer()
 
+        # Per-stream map of per-event toolCalls position -> logical tool-call
+        # index. GPT models stream parallel tool calls sequentially at
+        # position 0 (id+name opens a call, then id-less argument fragments
+        # follow), so a position can point at different logical calls over the
+        # stream's lifetime. Reset by `reset_stream_state()`.
+        self._active_tool_call_indices: Dict[int, int] = {}
+
         # Chat request and message models
         self.oci_chat_request = models.GenericChatRequest
         self.oci_chat_message = {
@@ -288,6 +295,7 @@ class GenericProvider(Provider):
         caller-owned state via :meth:`new_stream_state`, which needs no reset.
         """
         self._xml_buffer.reset()
+        self._active_tool_call_indices.clear()
 
     def flush_stream_state(self) -> str:
         """Return held-back text from the legacy instance-level buffer.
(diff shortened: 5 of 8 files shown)


fede-kamel:
Folded in the reverse id→index lookup from #252 (`58fa3778`): if a provider re-announces an id-bearing chunk for an already-open call at a position occupied by a different call, the call now keeps its original logical index instead of splitting its arguments across two tool calls. Unit test added for the re-send shape.

Verified by a differential probe running this branch's algorithm and #252's side by side across six stream shapes (GPT parallel from #253, gpt-oss fragmented, Grok parallel, Gemini id-less, multi-item fragments, id re-send conflict): routing is identical and correct on all six, so this PR now strictly covers #252's behavior. Full unit suite (457 passed), lint/mypy, and the live integration tests all green after the change.

fede-kamel:
Differential probe evidence backing the "strictly covers #252" claim in the previous comment — both algorithms (#252's transcribed verbatim from its diff; this branch's via the real `GenericProvider`) were fed six stream shapes and their `(id, logical_index)` routing compared. Correctness criterion: no id split across two indexes, no index shared by two ids.

Shapes: GPT parallel (#253 repro), gpt-oss fragmented single call, Grok parallel, Gemini id-less, multi-item array-parallel fragments, id re-send at a conflicting position.

Before `58fa3778`:

```text
gpt-parallel (#253)      252:OK  254:OK  identical:True
gpt-oss single           252:OK  254:OK  identical:True
grok parallel            252:OK  254:OK  identical:True
gemini idless            252:OK  254:OK  (uuid labels differ only)
multi-item frags         252:OK  254:OK  identical:True
id re-send conflict      252:OK  254:BAD
   252: [('call_A', 0), ('call_B', 1), ('call_B', 1)]
   254: [('call_A', 0), ('call_B', 1), ('call_B', 2)]  <- call_B split
```

After `58fa3778` (reverse id→index lookup ported from #252):

```text
gpt-parallel (#253)      252:OK  254:OK  identical:True
gpt-oss single           252:OK  254:OK  identical:True
grok parallel            252:OK  254:OK  identical:True
gemini idless            252:OK  254:OK  (uuid labels differ only)
multi-item frags         252:OK  254:OK  identical:True
id re-send conflict      252:OK  254:OK  identical:True
```

All 12 CI checks green on the current head.

adoerfle (approved):
Thank you. I was able to test this patch against our GPT-4.1 use case and confirm it resolves the issue.
With the patch, the streamed parallel tool calls reconstruct correctly: all three tool calls keep their own populated args, and the final tool invocations receive the expected kwargs instead of {}.

fileames:
`_active_tool_call_indices` is described as per-stream state but is stored on the shared GenericProvider. Concurrent stream() or astream() calls on the same ChatOCIGenAI instance reset and overwrite each other’s routing map.
Example interleaving:

>   1. Stream A opens calls A and B at position 0; its active index becomes 1.
>   2. Stream B starts, clears the shared map, and sets position 0 to index 0.
>   3. Stream A receives B’s next id-less argument fragment.
>   4. The fragment is routed to index 0 and therefore attached to call A.

This silently corrupts tool-call arguments in normal concurrent server or LangChain batch usage. Passing tests only validate isolated streams. Do we expect concurrent streams?

fede-kamel:
Good catch — confirmed. `_active_tool_call_indices` was provider-instance state, so the interleaving you describe would route stream A's id-less fragment to the wrong call.

To answer the question: yes, we expect concurrent streams — this is the same class of bug as #255 (concurrent streams sharing the GenericProvider's XML `<tool_call>` buffer), which #256 fixes for that buffer by making the state caller-owned.

Fixed in b379e96e using the pattern this file already uses for `tool_call_ids`: the routing map is now created per-stream in `_stream`/`_astream` and threaded through `process_stream_tool_calls(event_data, tool_call_ids, active_tool_call_indices)`; nothing is stored on (or reset via) the shared provider. Added `test_process_stream_tool_calls_routing_state_is_per_stream`, which reproduces your exact 4-step interleaving and asserts the fragment stays attached to call B at index 1. Full unit suite green (472 passed).

fileames (dismissed):

fede-kamel:
Pre-merge live e2e against OCI GenAI (us-chicago-1) using the newly-listed GPT 5.6 models — `openai.gpt-5.6-sol`, `openai.gpt-5.6-luna`, `openai.gpt-5.6-terra` (each with `reasoning_effort: NONE`, which the 5.6 line requires for function tools on the Chat Completions path).

Two scenarios per model, differential against pre-fix `main` (identical script, real streamed parallel tool calls: get_weather + get_population for two cities = 4 parallel calls):

**1. Sync `stream()` parallel tool calls (issue #253):**
- `main`: **FAIL on all 3 models** — only the first tool call keeps its arguments; the other three reconstruct with empty `{}` args (the exact #253 symptom).
- this branch: **PASS on all 3 models** — all 4 calls reconstruct with their own complete args.

**2. Concurrent `astream()` on one shared `ChatOCIGenAI` instance (@fileames' scenario), two interleaved streams with disjoint city sets:**
- this branch: **PASS on all 3 models** — each stream's 4 tool calls come back intact, valid JSON, and every city stays in its own stream (no cross-stream contamination, no duplicate ids).

```text
                         sync-parallel   concurrent-astream
openai.gpt-5.6-sol       PASS            PASS
openai.gpt-5.6-luna      PASS            PASS
openai.gpt-5.6-terra     PASS            PASS
(main @ ead5888a)        FAIL ({} args)  FAIL ({} args)
```

Combined with the green unit suite (472) and CI (12/12), this is ready to merge from my side.

fede-kamel:
@fileames the rebase onto #256's merge dismissed your approval — could you re-approve? Changes since your review: conflict resolution only, combining the two per-stream mechanisms in `process_stream_tool_calls(event_data, tool_call_ids, stream_state=…, active_tool_call_indices=…)` (both caller-owned). No logic changes beyond that. Full unit suite (486) green, 12/12 CI checks green, and I re-ran the live e2e (3× GPT 5.6 models, sync parallel + concurrent astream) on the combined code — all pass.

fileames (approved):

PR #280, opened 2026-07-22
feat(oracledb): support nested metadata filters in OracleVS

### LINKED ISSUE #272: feat(oracledb): support nested metadata filters in OracleVS ($and/$or composition)
## Motivation

`OracleVS` metadata filtering supports a solid operator set (`$and`/`$or`, `$eq`, `$between`, comparison operators), but **nested filter structures raise `ValueError`** (`libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py`, `_generate_where_clause` — "nested filters unsupported"). Users cannot express queries like `{"$and": [{"$or": [...]}, {...}]}` that peer stores (PGVector/PGVectorStore, MongoDB Atlas, Elasticsearch) handle routinely. This also caps the operator coverage achievable by a future self-query translator (tracked separately).

## Proposed scope

- Make `_generate_condition` / `_generate_where_clause` recursive so `$and`/`$or` compose to arbitrary depth, generating parenthesized SQL over the JSON metadata column.
- Keep bind-variable-based value passing (no SQL injection surface).
- Same treatment in the JS `OracleVS` filter builder (`libs/js/langchain-oracledb/src/vectorstores.ts`) if it shares the limitation.

## Acceptance criteria

- Arbitrarily nested `$and`/`$or` combinations produce correct results (unit tests on generated SQL + integration tests on result sets).
- Existing flat-filter behavior unchanged.

fede-kamel:
## Summary

Implements #272. One scoping correction discovered during implementation: nested **logical** composition (`{"$and": [{"$or": [...]}, ...]}`) already worked — `_generate_where_clause` recurses over `$and`/`$or`/`$nor`. The actual limitation was nested **metadata paths**: `{"address": {"city": "NYC"}}` raised `ValueError("Nested filters are not supported.")`.

Nested dicts are now translated recursively into dotted JSON paths (`$.address.city`), exactly equivalent to the already-supported `{"address.city": "NYC"}` spelling:

```python
vs.similarity_search(q, filter={"address": {"city": "NYC", "zip": {"$ne": "10001"}}})
vs.similarity_search(q, filter={"$or": [{"address": {"city": "LA"}}, {"level": {"$gte": 3}}]})
```

- Operator dicts and scalars terminate the recursion → arbitrary nesting depth
- Sibling keys at the same nesting level combine with `AND`
- Values still flow through positional bind variables (no SQL injection surface)
- Mixing operator keys with nested keys in one dict (`{"a": {"$eq": 1, "b": 2}}`) is ambiguous and now fails with a specific message instead of the generic error

## Files changed

- @libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py — recursive nested-path branch in `_generate_condition`
- @libs/oracledb/tests/unit_tests/vectorstores/test_nested_filters.py — new: 9 tests pinning the emitted `JSON_EXISTS` SQL and bind lists, including nested/dotted equivalence and the pre-existing logical recursion
- @libs/oracledb/tests/integration_tests/vectorstores/test_nested_filters.py — new: 5 env-gated tests (`VECDB_*` convention)

## Verification

- Unit: 385 passed (9 new); ruff + mypy clean.
- Live against Oracle Database 23ai Free: all 5 integration tests pass — nested scalar path, nested siblings + `$ne`, `$or` over a nested branch, `$and`+`$or` logical nesting combined with a nested path, and nested-vs-dotted result equivalence.

Closes #272

### DIFF
--- a/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
+++ b/libs/oracledb/langchain_oracledb/vectorstores/oraclevs.py
@@ -349,8 +349,30 @@ def _generate_condition(
 
             return res
 
+        elif any(value_key.startswith("$") for value_key in value.keys()):
+            raise ValueError(
+                "Invalid filter: operator keys ($...) cannot be mixed with "
+                f"nested metadata keys in {value!r}."
+            )
+
         else:
-            raise ValueError("Nested filters are not supported.")
+            # Nested metadata path: {"address": {"city": "NYC"}} is
+            # equivalent to {"address.city": "NYC"}. Recurse with the
+            # dotted path; operator dicts and scalars terminate the
+            # recursion, so arbitrary nesting depth is supported.
+            nested_conditions = [
+                _generate_condition(
+                    f"{metadata_key}.{sub_key}", sub_value, bind_variables
+                )
+                for sub_key, sub_value in value.items()
+            ]
+
+            res = " AND ".join(nested_conditions)
+
+            if len(nested_conditions) > 1:
+                return "(" + res + ")"
+
+            return res
 
     else:
         raise ValueError("Filter format is invalid.")
--- a/libs/oracledb/tests/integration_tests/vectorstores/test_nested_filters.py
+++ b/libs/oracledb/tests/integration_tests/vectorstores/test_nested_filters.py
@@ -1,0 +1,112 @@
+# Copyright (c) 2026 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Integration tests for nested metadata filters in OracleVS (issue #272).
+
+Requires a reachable Oracle Database 23ai+ via the usual env vars:
+VECDB_USER, VECDB_PASS, VECDB_HOST (dsn).
+"""
+
+from __future__ import annotations
+
+import os
+from typing import Any, List
+
+import oracledb
+import pytest
+from langchain_core.documents import Document
+from langchain_core.embeddings import DeterministicFakeEmbedding
+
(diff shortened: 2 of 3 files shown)


fileames:
Thanks a lot for this.
Could we apply the same nested-path approach to the LangGraph Oracle checkpoint saver as well, or should we create separate issue and PR tracking that?

In `libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/base.py`, `_search_where()` currently uses `JSON_EQUAL` for dictionaries and lists. This requires an exact match, while the LangGraph.js Oracle implementation uses recursive containment. For example, filtering with {"nested": {"b": {"c": true}}} should match metadata containing {"nested": {"a": 1, "b": {"c": true}}}.

fede-kamel:
@fileames when you get a chance, could you review this one? It implements #272 (nested $and/$or metadata filter composition in OracleVS) — CI is green and it's been ready since last week. Thanks!

fede-kamel:
@fileames sorry for the delayed reply — good catch, and confirmed: `_search_where()` in the checkpoint saver uses `JSON_EQUAL` on `JSON_QUERY(metadata, '$.key')` for dict/list values, so `{"nested": {"b": {"c": true}}}` will not match `{"nested": {"a": 1, "b": {"c": true}}}`, which diverges from both the LangGraph.js Oracle saver and PostgresSaver's containment semantics.

I'd track it separately: it's a different package (`langgraph-oracledb`, its own version/release) and a different filter surface (checkpoint metadata containment vs. vector-store filter operators), so bundling it here would couple two releases. Filed #290 to track it — the plan is to reuse this PR's nested-path flattening approach there.

This PR stays scoped to OracleVS as-is, so it's ready from my side.

fileames (approved):

PR #277, opened 2026-07-22
test(oci): adopt langchain-tests standard ChatModel conformance suite

fede-kamel:
## Summary

Implements #261: adds the official `langchain-tests` conformance suite for the chat models in `libs/oci`, plus the library fixes for every gap the suite surfaced.

**New tests**
- `tests/unit_tests/chat_models/test_standard.py` — `ChatModelUnitTests` for `ChatOCIGenAI` on the Meta and Cohere provider paths (capability flags set honestly, e.g. `has_tool_choice=False` for Cohere per #266) and for `ChatOCIModelDeployment` (gated on `ads`).
- `tests/integration_tests/chat_models/test_standard.py` — `ChatModelIntegrationTests` for `ChatOCIGenAI`, gated on `OCI_COMPARTMENT_ID` following the existing integration-test conventions.

**Conformance fixes surfaced by the suite**
- `ChatOCIGenAI` now accepts standard top-level params: `temperature`, `max_tokens`, `stop`, and `streaming`. `streaming` is a constructor alias for `is_stream` implemented as a before-validator rather than a pydantic field — the `@pre_init` validator marks every field as explicitly set, and langchain-core's `_streaming_disabled` would then treat a defaulted `streaming=False` as an explicit opt-out and silently reroute every `.stream()` call to `.invoke()`.
- `_get_ls_params` now emits `ls_model_name` / `ls_temperature` / `ls_max_tokens` / `ls_stop`, so LangSmith traces attribute calls to the correct model.
- `response_metadata` now includes the standard `model_name` key (invoke and final stream chunk); `model_id` is kept for backwards compatibility.
- `with_structured_output` accepts the standard `strict` kwarg (mapped to the `json_schema` `is_strict` flag) and binds `ls_structured_output_format` metadata for tracing callbacks.
- `tool_choice` validation is deferred from bind time to request time, so binding stays declarative; providers that reject `tool_choice` (Cohere) now raise on invoke, not on bind.
- Pydantic **v1** model classes are now recognized by `with_structured_output` (`is_pydantic_class` uses `is_basemodel_subclass`), so results parse to the model instance instead of a dict.
- The Generic provider path now skips `tool_use`/`tool_call` content blocks in list-form `AIMessage` histories instead of raising — the calls are transmitted through the request's dedicated `tool_calls` field. This fixes both `test_tool_message_histories_list_content` and the `create_agent` agent-loop test.
- `model_kwargs` is copied before adding per-request stop sequences, fixing a latent shared-dict mutation.

[...]

Issue #260, opened 2026-07-22
feat(oci): support standard content blocks and output_version="v1" in ChatOCIGenAI (LangChain 1.0)


fede-kamel:
## Motivation

LangChain 1.0 introduced **standard content blocks** — a unified, typed `content_blocks` representation on messages (text, reasoning, citations, tool calls, server-side tool results) — plus the `output_version="v1"` flag that makes `AIMessage.content` the standardized list-of-blocks serialization, including in streaming (`stream_events` v3 protocol). This is the bar for first-class provider chat models: langchain-openai, langchain-anthropic, langchain-aws, langchain-google, and langchain-ollama all implement it.

See: https://docs.langchain.com/oss/python/releases/langchain-v1

## Current state

- `ChatOCIGenAI` (`libs/oci/langchain_oci/chat_models/oci_generative_ai.py`) has no content-block or `output_version` support — responses are emitted in provider-native shapes only.
- `ChatOCIOpenAI` gets v1 support for free by inheriting `langchain_openai.ChatOpenAI` (it already plumbs `output_version` for the Responses API).
- `ChatOCIModelDeployment` likewise has no support.

## Proposed scope

- Translate OCI GenAI native response events (across the Cohere, Generic/Meta, Gemini, and OpenAI provider paths) into standard content blocks.
- Support `output_version="v1"` on `ChatOCIGenAI`, including v1-format chunks during streaming.
- Cover reasoning content for reasoning-capable models where the service surfaces it.

## Acceptance criteria

- `ChatOCIGenAI(..., output_version="v1")` emits list-of-content-blocks `AIMessage.content` for invoke and stream.
- `message.content_blocks` returns typed blocks regardless of `output_version`.
- The `langchain-tests` `test_stream_events_v3` standard tests pass (see companion issue on adopting the standard test suite).

PR #281, opened 2026-07-22
feat(oci): standard content blocks and output_version="v1" support in ChatOCIGenAI

fede-kamel:
## Summary

Implements #260 — LangChain 1.0 standard content blocks for `ChatOCIGenAI`.

Key finding that shaped the implementation: **langchain-core ≥ 1.0 already handles the output side.** With `output_version="v1"` the base model rewrites `AIMessage.content` into the standardized list-of-blocks form, and core's parser maps our message shape (text, `additional_kwargs["reasoning_content"]`, `tool_calls`) to `text`/`reasoning`/`tool_call` blocks. What was broken was the **input side**: feeding a v1-format message back into the model raised `ValueError: Unsupported content type: reasoning` on the Generic/Meta path, and the Cohere V1 path would push list content into string-typed request fields — which made any multi-turn v1 conversation or agent loop impossible.

**Changes**
- `GenericProvider._process_message_content` skips `reasoning`/`tool_call`/`tool_use` blocks — tool calls travel via `AIMessage.tool_calls`, and reasoning is model output the API doesn't accept back as input
- New `OCIUtils.content_to_text` (joins text blocks in order, skips dedicated-field block types); the Cohere V1 path uses it at every string-content site (history, AIMessage content, dummy repeat message, final message). The Cohere V2 vision path already tolerated unknown blocks
- 6 new unit tests covering the coercion helper, both provider request builders with v1 histories, and a mocked end-to-end v1 invoke

```python
llm = ChatOCIGenAI(model_id=..., output_version="v1")
llm.invoke("hi").content
# [{'type': 'text', 'text': 'Hello!'}]
```

## Files changed

- @libs/oci/langchain_oci/chat_models/providers/generic.py — skip v1 dedicated-field blocks in content parsing
- @libs/oci/langchain_oci/chat_models/providers/cohere.py — coerce content via `content_to_text` at V1 string sites
- @libs/oci/langchain_oci/common/utils.py — `OCIUtils.content_to_text`
- @libs/oci/tests/unit_tests/chat_models/test_content_blocks_v1.py — new

## Verification

- Unit: 492 passed (6 new); ruff + mypy clean.
- Live against OCI GenAI, **7/7**: v1 invoke returns block lists; v1 streaming chunks aggregate correctly; v1 histories containing `reasoning`+`tool_call` blocks round-trip; a complete tool-calling round-trip runs entirely under v1 on both Meta and Cohere paths; **`xai.grok-3-mini` emits a standard `reasoning` block alongside `text` under v1**; Cohere accepts v1 histories.

Note: the Generic-provider hunk touches the same region as #277's `tool_use`/`tool_call` skip (this one supersedes it by adding `reasoning`); whichever merges second is a one-hunk rebase.

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/cohere.py
+++ b/libs/oci/langchain_oci/chat_models/providers/cohere.py
@@ -496,7 +496,9 @@ class CohereProvider(Provider):
             role = self.get_role(msg)
             if role in ("USER", "SYSTEM"):
                 oci_chat_history.append(
-                    self.oci_chat_message[role](message=msg.content)
+                    self.oci_chat_message[role](
+                        message=OCIUtils.content_to_text(msg.content)
+                    )
                 )
             elif isinstance(msg, AIMessage):
                 # Skip tool calls if forcing single step
@@ -510,7 +512,7 @@ class CohereProvider(Provider):
                     if msg.tool_calls
                     else None
                 )
-                msg_content = msg.content if msg.content else " "
+                msg_content = OCIUtils.content_to_text(msg.content) or " "
                 oci_chat_history.append(
                     self.oci_chat_message[role](
                         message=msg_content, tool_calls=tool_calls
@@ -541,7 +543,9 @@ class CohereProvider(Provider):
                     # by an AI message
                     oci_chat_history.append(
                         self.oci_chat_message["CHATBOT"](
-                            message=messages[len(messages) - i - 2].content
+                            message=OCIUtils.content_to_text(
+                                messages[len(messages) - i - 2].content
+                            )
                         )
                     )
                 break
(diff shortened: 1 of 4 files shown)


fileames:
For `content_to_text`, should we raise when expected type doesn't match? Just as a suggestion otherwise looks good, approving, thanks :)

fileames (approved):

fede-kamel:
Thanks for the review! I went back and forth on this one, but I'd keep it lenient — raising on unexpected blocks is basically the bug this PR fixes. The Generic path used to throw `Unsupported content type: reasoning`, which killed any multi-turn v1 conversation. Non-text blocks get skipped on purpose since they already travel through their own request fields (`tool_calls` etc.), so nothing is actually lost. It also keeps us from breaking when core adds new block types down the road — same approach the Cohere V2 vision path takes. That said, if a case comes up where silent skipping hides a real problem, adding a strict mode would be easy.

Issue #276, opened 2026-07-22
feat(js): OCI Generative AI chat model and embeddings for LangChain.js


fede-kamel:
## Motivation

There is no LangChain.js equivalent of `langchain-oci`: no OCI GenAI chat model, embeddings, or completion LLM for the JavaScript/TypeScript ecosystem. JS/TS is a major share of the LangChain user base (LangChain.js, LangGraph.js), and today those users cannot call OCI Generative AI through a first-party integration — the community `@langchain/community` package has no maintained OCI GenAI chat model either.

Peer precedent: every major provider ships a JS package alongside Python (`@langchain/openai`, `@langchain/aws`, `@langchain/google-vertexai`).

## Proposed scope (phased)

1. **`ChatOCIGenAI`** (extends LangChain.js `BaseChatModel`): invoke + streaming, tool calling, structured output, usage metadata, using the `oci-sdk` for JS/TS (or signed fetch) with the same auth modes as Python (`API_KEY`, `SECURITY_TOKEN`, instance/resource principals).
2. **`OCIGenAIEmbeddings`**.
3. Later: OpenAI-compatible path for OCI-hosted OpenAI models (JS analog of `ChatOCIOpenAI`, potentially a thin wrapper over `@langchain/openai` with OCI request signing).

Packaging question to settle first: extend `@oracle/langchain-oracledb`'s sibling directory with a new `@oracle/langchain-oci` package under `libs/js/`.

## Acceptance criteria

- `ChatOCIGenAI` passes the LangChain.js standard chat-model tests.
- Works in LangGraph.js agent loops (tool round-trips).
- Published to npm under the `@oracle` scope.

sharadraju:
@fede-kamel Can we close this issue, now that @oracle/langchain-oci is released

fede-kamel:
Yes, let's close it. `@oracle/langchain-oci` covers the first two phases: `OciGenAiGenericChat` and `OciGenAiCohereChat` with invoke, streaming, tool calling and structured output, `OciGenAiEmbeddings`, API-key and session-token auth, published under the `@oracle` scope (0.0.1 out, 0.0.2 on its way with the ESM/CJS declaration fix from #312). Two items from the original scope are not in yet and can get their own issues if we want them: the OpenAI-compatible path for OCI-hosted OpenAI models (phase 3) and adopting the LangChain.js standard chat-model test suite.

PR #285, opened 2026-07-30
Add OCI GenAI integration for TS

sudarshan12s:
Basic implementation for OCI Generartive AI integration with LangChainJS.

## unit test

cd libs/js/oci
pnpm install
pnpm build   
pnpm test   
pnpm run lint
pnpm vitest run src/tests/chatoci_genai.test.ts
pnpm vitest run src/tests/langgraph_tool_roundtrip.test.ts

## integration tests

export OCI_COMPARTMENT_ID=<id>
export OCI_CONFIG_FILE="$HOME/.oci/config"
export OCI_CONFIG_PROFILE="DEFAULT"
export OCI_REGION="us-phoenix-1"
export OCI_ENDPOINT="https://inference.generativeai.us-phoenix-1.oci.oraclecloud.com"

export OCI_GENAI_INTEGRATION_TESTS_COMPARTMENT_ID="$OCI_COMPARTMENT_ID"
export OCI_GENAI_INTEGRATION_TESTS_GENERIC_ON_DEMAND_MODEL_ID="xai.grok-3"
export OCI_GENAI_INTEGRATION_TESTS_CHAT_MODELS="generic"
export OCI_MODEL_ID="xai.grok-3"

pnpm vitest run --mode int src/tests/chatoci_genai.int.test.ts
pnpm test:langgraph:int


libs/js/oci/src/server_events_iterator.ts
```diff
@@ -0,0 +1,68 @@
...
+  _textDecoder: TextDecoder = new TextDecoder();
+
+  _textBuffer: string = "";
+
+  constructor(sourceStream: ReadableStream<Uint8Array>) {
+    this._eventsStream =
+      IterableReadableStream.fromReadableStream(sourceStream);
+  }
+
+  async *[Symbol.asyncIterator](): AsyncIterator<unknown> {
```
fede-kamel, line 27: **Blocker — this breaks when one network chunk carries more than one SSE event.** The buffer is parsed as a *single* `data: ` message once it ends with `\n\n`, but TCP coalescing routinely delivers `data: {...}\n\ndata: {...}\n\n` in a single read. In that case `_parseMessage()` strips the first prefix and calls `JSON.parse` on `{...}\n\ndata: {...}` → throws `Could not parse event data as JSON` and kills the stream mid-response. The faster the model streams, the more likely this is to happen.

The iterator needs to split the buffer on the event delimiter and parse each event separately, keeping any trailing partial event in the buffer, e.g.:

```ts
this._textBuffer += this._textDecoder.decode(eventRawData, { stream: true });
const events = this._textBuffer.split("\n\n");
this._textBuffer = events.pop() ?? ""; // partial tail stays buffered
for (const event of events) {
  if (event.trim().length === 0) continue;
  yield this._parseSingleEvent(event);
}
```

A unit test with two events in one chunk (and one event split across two chunks) would lock this in.

libs/js/oci/src/server_events_iterator.ts
```diff
@@ -0,0 +1,68 @@
...
+    for await (const eventRawData of this._eventsStream) {
+      this._textBuffer += this._textDecoder.decode(eventRawData);
```
fede-kamel: `decode(eventRawData)` without `{ stream: true }` corrupts multi-byte UTF-8 characters that happen to be split across two network chunks (emoji, accented text) — each partial sequence becomes U+FFFD. Use `this._textDecoder.decode(eventRawData, { stream: true })`.

libs/js/oci/src/generic_chat.ts
```diff
@@ -0,0 +1,193 @@
...
+      typeof content === "object" &&
+      (<TextContent>content).type === TextContent.type &&
+      typeof (<TextContent>content).text === "string"
+    );
+  }
+
+  static _getChunkDataText(chunkData: ChatChoice): string | undefined {
+    return chunkData.message?.content
+      ?.map((message: TextContent) => message.text)
+      .join(" ");
```
fede-kamel: Streamed chunks join content parts with `" "` while the non-streaming `_parseResponse` joins with `""` — so `stream()` and `invoke()` can produce different text for the same completion whenever a chunk carries multiple content parts. These should agree (join with `""`).

libs/js/oci/src/cohere_chat.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  override _parseStreamedResponseChunk(chunk: unknown): string {
+    if (OciGenAiCohereChat._isCohereChunkData(chunk)) {
+      return chunk.text;
+    }
+
+    throw new Error("Invalid streamed response chunk data");
+  }
+
+  static _splitMessageAndHistory(messages: BaseMessage[]): HistoryMessageInfo {
```
fede-kamel, line 84: Two edge cases here:

1. If the conversation ends with an AI/system message (or the only user turn is mid-history), the *last user message anywhere* is promoted to `message` and everything after it — including AI turns — stays in `chatHistory`. That silently reorders the conversation the model sees.
2. If there is no human message at all, `message` is `""`, which the API will reject with an opaque 400.

Suggest requiring the final message to be a human message (throw a clear error otherwise), which matches how the Cohere apiFormat is meant to be used.

libs/js/oci/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "lint:fix": "pnpm lint:eslint --fix && pnpm lint:dpdm",
+    "clean": "rm -rf dist/ dist-types-cjs && node scripts/create-entrypoints.js pre",
+    "prepack": "pnpm build",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "test:int": "vitest run --mode int",
+    "format": "prettier --config .prettierrc --write \"src\"",
+    "format:check": "prettier --config .prettierrc --check \"src\""
+  },
+  "author": "LangChain",
```
fede-kamel: `"author": "LangChain"` — should be Oracle for an `@oracle/*` package in this repo.

fede-kamel (changes requested):
Thanks for this contribution — a TS OCI GenAI chat integration is very welcome (tracked in #276), and the unit test coverage (1400+ lines) is great to see. Overall structure is clean and the auth/client separation is nice.

Requesting changes for one functional blocker in the SSE stream parsing, plus a few smaller items — details inline.

A few non-blocking, bigger-picture notes:

1. **Text-only + `SimpleChatModel`**: the integration rejects non-string content and doesn't surface tool calls, token usage, or `finish_reason`. The Python `ChatOCIGenAI` supports tool calling, structured output, and multimodal content, and #276 asks for parity including embeddings. It's fine to land chat-text-first, but consider building on `BaseChatModel` and returning a proper `ChatResult` with `usage_metadata`/`response_metadata` now, so tool calling can be added without a breaking rewrite. Could you say what the intended follow-up roadmap is (tools, embeddings, `OciGenAiEmbeddings`)?
2. **Auth parity**: `OciGenAiNewClientAuthType` covers config file, instance principal, session, and custom — Resource Principal (supported by the Python lib and needed for OCI Functions/Data Science) would be a cheap addition since `oci-common` ships a provider for it.
3. **CI**: nothing in the repo CI builds or tests `libs/js/oci`, so the checks passing on this PR doesn't exercise any of this code. Worth wiring the unit tests into CI in this PR if possible.
4. `_llmType()` returns `"custom"` — should be something like `"oci_genai"` so runs are identifiable in tracing/LangSmith.

fede-kamel:
Following up on my review with an executable repro of the two `JsonServerEventsIterator` findings — both confirmed with this unit test (drop it in `src/tests/` and run `pnpm vitest run src/tests/repro_sse.test.ts`):

```ts
import { expect, test } from "vitest";
import { JsonServerEventsIterator } from "../server_events_iterator.js";

function streamOf(chunks: Uint8Array[]): ReadableStream<Uint8Array> {
  return new ReadableStream<Uint8Array>({
    start(controller) {
      for (const chunk of chunks) controller.enqueue(chunk);
      controller.close();
    },
  });
}

const enc = new TextEncoder();

test("two SSE events coalesced into one chunk should both parse", async () => {
  const chunk = enc.encode('data: {"text":"hello"}\n\ndata: {"text":"world"}\n\n');
  const events: unknown[] = [];
  for await (const e of new JsonServerEventsIterator(streamOf([chunk]))) events.push(e);
  expect(events).toEqual([{ text: "hello" }, { text: "world" }]);
});

test("multi-byte UTF-8 split across two chunks should decode intact", async () => {
  const whole = enc.encode('data: {"text":"a🙂b"}\n\n');
  const splitAt = whole.indexOf(0xf0) + 2; // mid-emoji
  const events: unknown[] = [];
  for await (const e of new JsonServerEventsIterator(streamOf([whole.slice(0, splitAt), whole.slice(splitAt)]))) events.push(e);
  expect(events).toEqual([{ text: "a🙂b" }]);
});
```

Results on the current branch (`70b543a2`):

```
× two SSE events coalesced into one chunk should both parse
  → Could not parse event data as JSON          # stream dies mid-response
× multi-byte UTF-8 split across two chunks should decode intact
  → expected [ { text: 'a���b' } ] to deeply equal [ { text: 'a🙂b' } ]
```

Both failures match the inline comments: the buffer is parsed as a single `data:` message, and `TextDecoder.decode()` is called without `{ stream: true }`.

I also verified the fix suggested in the review (decode with `{ stream: true }`, split the buffer on `\n\n`, keep the partial tail buffered, parse each event separately): with that change applied to `server_events_iterator.ts`, **both repro tests pass and all 31 existing unit tests in `chatoci_genai.test.ts` still pass**. The two tests above would make good permanent regression tests once the fix lands.

libs/js/oci/src/server_events_iterator.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      if (this._completeMessageReceived()) {
+        yield this._parseMessage();
+        this._textBuffer = "";
+      }
+    }
+  }
```
fede-kamel, line 72:
```suggestion
  async *[Symbol.asyncIterator](): AsyncIterator<unknown> {
    for await (const eventRawData of this._eventsStream) {
      this._textBuffer += this._textDecoder.decode(eventRawData, {
        stream: true,
      });

      const events = this._textBuffer.split("\n\n");
      this._textBuffer = events.pop() ?? "";
      for (const event of events) {
        if (event.trim().length === 0) continue;
        yield this._parseSingleEvent(event);
      }
    }
  }

  _parseSingleEvent(eventText: string): unknown {
    if (!eventText.startsWith(JsonServerEventsIterator._DATA_PREFIX)) {
      throw new Error("Event text is empty, too short or malformed");
    }
    return this._tryParseTextToJson(
      eventText.substring(JsonServerEventsIterator._DATA_PREFIX_LENGTH)
    );
  }
```
Handles multiple SSE events per network chunk (split on `\n\n`, partial tail stays buffered) and decodes with `{ stream: true }` so multi-byte UTF-8 split across chunks stays intact. `_completeMessageReceived`/`_parseMessage`/`_assertDataMessage` become unused and can be deleted along with their tests.

libs/js/oci/src/generic_chat.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
```
fede-kamel, line 391:
```suggestion
  static _getChunkDataText(chunkData: ChatChoice): string | undefined {
    return chunkData.message?.content
      ?.map((message: TextContent) => message.text)
      .join("");
  }
```
Aligns streamed chunk concatenation with `_parseResponse` (empty-string join) so `stream()` and `invoke()` produce identical text.

libs/js/oci/src/chat_models.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+  }
+
+  static _isValidString(value: unknown): value is string {
+    return typeof value === "string" && value.length > 0;
+  }
+
+  _llmType() {
+    return "custom";
+  }
```
fede-kamel, line 484:
```suggestion
  _llmType() {
    return "oci_genai";
  }
```

fede-kamel (commented):
To make the fixes easy to land, here are the mechanical ones as one-click suggestions. The SSE iterator suggestion is exactly the variant I verified locally: both repro tests pass and all 31 existing unit tests in `chatoci_genai.test.ts` stay green with it applied.

libs/js/oci/package.json
fede-kamel:
```suggestion
  "author": "Oracle",
```

sudarshan12s:
> Following up on my review with an executable repro of the two `JsonServerEventsIterator` findings — both confirmed with this unit test (drop it in `src/tests/` and run `pnpm vitest run src/tests/repro_sse.test.ts`):
> 
> ```ts
> import { expect, test } from "vitest";
> import { JsonServerEventsIterator } from "../server_events_iterator.js";
> 
> function streamOf(chunks: Uint8Array[]): ReadableStream<Uint8Array> {
>   return new ReadableStream<Uint8Array>({
>     start(controller) {
>       for (const chunk of chunks) controller.enqueue(chunk);
>       controller.close();
>     },
>   });
> }
> 
> const enc = new TextEncoder();
> 
> test("two SSE events coalesced into one chunk should both parse", async () => {
>   const chunk = enc.encode('data: {"text":"hello"}\n\ndata: {"text":"world"}\n\n');
>   const events: unknown[] = [];
>   for await (const e of new JsonServerEventsIterator(streamOf([chunk]))) events.push(e);
>   expect(events).toEqual([{ text: "hello" }, { text: "world" }]);
> });
> 
> test("multi-byte UTF-8 split across two chunks should decode intact", async () => {
>   const whole = enc.encode('data: {"text":"a🙂b"}\n\n');
>   const splitAt = whole.indexOf(0xf0) + 2; // mid-emoji
>   const events: unknown[] = [];
>   for await (const e of new JsonServerEventsIterator(streamOf([whole.slice(0, splitAt), whole.slice(splitAt)]))) events.push(e);
>   expect(events).toEqual([{ text: "a🙂b" }]);
> });
> ```
> 
> Results on the current branch (`70b543a2`):
> 
> ```
> × two SSE events coalesced into one chunk should both parse
>   → Could not parse event data as JSON          # stream dies mid-response
> × multi-byte UTF-8 split across two chunks should decode intact
>   → expected [ { text: 'a���b' } ] to deeply equal [ { text: 'a🙂b' } ]
> ```
> 
> Both failures match the inline comments: the buffer is parsed as a single `data:` message, and `TextDecoder.decode()` is called without `{ stream: true }`.
> 
> I also verified the fix suggested in the review (decode with `{ stream: true }`, split the buffer on `\n\n`, keep the partial tail buffered, parse each event separately): with that change applied to `server_events_iterator.ts`, **both repro tests pass and all 31 existing unit tests in `chatoci_genai.test.ts` still pass**. The two tests above would make good permanent regression tests once the fix lands.

Thanks @fede-kamel for your thoughtful and insightful comments! The initial changes are quite old and were originally authored by Yanir Shahak. I’m working on resuming the effort and bringing it up to date with the latest SDK changes. I’m still getting familiar with the code and overall flow, but I’m hoping to have it ready for review by the end of this week.

libs/js/oci/src/chat_models.ts
sudarshan12s: Done

libs/js/oci/src/cohere_chat.ts
sudarshan12s: `_splitMessageAndHistory` now requires final element in messages to be a human message and throw a error otherwise.

libs/js/oci/src/generic_chat.ts
sudarshan12s: Done

libs/js/oci/src/server_events_iterator.ts
sudarshan12s: Fixed it. also handled `CR (\r), \r\n\r\n `  `JsonServerEventsIterator`* tests cover the scenarios.

libs/js/oci/src/server_events_iterator.ts
sudarshan12s: Thanks. Now it handles multi byte utf8 split . Added tests  (`preserves UTF-8 characters split acorss chunks`) for the same.

libs/js/oci/src/server_events_iterator.ts
sudarshan12s: Done.

libs/js/oci/src/generic_chat.ts
sudarshan12s: Done.

libs/js/oci/package.json
sudarshan12s: Done

sudarshan12s: Done

sudarshan12s:
> Thanks for this contribution — a TS OCI GenAI chat integration is very welcome (tracked in #276), and the unit test coverage (1400+ lines) is great to see. Overall structure is clean and the auth/client separation is nice.
> 
> Requesting changes for one functional blocker in the SSE stream parsing, plus a few smaller items — details inline.
> 
> A few non-blocking, bigger-picture notes:
> 
> 1. **Text-only + `SimpleChatModel`**: the integration rejects non-string content and doesn't surface tool calls, token usage, or `finish_reason`. The Python `ChatOCIGenAI` supports tool calling, structured output, and multimodal content, and #276 asks for parity including embeddings. It's fine to land chat-text-first, but consider building on `BaseChatModel` and returning a proper `ChatResult` with `usage_metadata`/`response_metadata` now, so tool calling can be added without a breaking rewrite. Could you say what the intended follow-up roadmap is (tools, embeddings, `OciGenAiEmbeddings`)?
> 2. **Auth parity**: `OciGenAiNewClientAuthType` covers config file, instance principal, session, and custom — Resource Principal (supported by the Python lib and needed for OCI Functions/Data Science) would be a cheap addition since `oci-common` ships a provider for it.
> 3. **CI**: nothing in the repo CI builds or tests `libs/js/oci`, so the checks passing on this PR doesn't exercise any of this code. Worth wiring the unit tests into CI in this PR if possible.
> 4. `_llmType()` returns `"custom"` — should be something like `"oci_genai"` so runs are identifiable in tracing/LangSmith.

- I have extended  `BaseChatModel` instead of  `SimpleChatModel `. Yes only chat-text is added, an error is thrown otherwise. I have added Tool Calling support and `OciGenEmbeddings`. I am not able to test `OciGenEmbeddings` , getting some error for my tenancy. I will try to use different setup and test it. 

- `OciGenAiNewClientAuthType` i used now.

- Added CI for js/oci. Currently CI fails in python files,  I see this patch could resolve it


```
- diff --git a/libs/oci/tests/unit_tests/chat_models/test_oci_responses_api.py b/libs/oci/tests/unit_tests/chat_models/test_oci_responses_api.py
index b159df7..96908ea 100644
--- a/libs/oci/tests/unit_tests/chat_models/test_oci_responses_api.py
+++ b/libs/oci/tests/unit_tests/chat_models/test_oci_responses_api.py
@@ -212,8 +212,8 @@ class TestResponseParsing:
 
     def test_empty_output(self, chat_model: Any) -> None:
         """Empty output should produce empty text, not crash."""
-        data = {"output": [], "usage": {}}
-        headers = {}
+        data: dict[str, Any] = {"output": [], "usage": {}}
+        headers: dict[str, str] = {}
 
         result = chat_model._process_responses_api_response(data, headers)
         assert result.data.chat_response.choices[0].message.content[0].text == ""
@@ -278,7 +278,7 @@ class TestResponseParsing:
             ],
             "usage": {"input_tokens": 5, "output_tokens": 4},
         }
-        headers = {}
+        headers: dict[str, str] = {}
 
         result = chat_model._process_responses_api_response(data, headers)
         assert (
```


- `_llmType()` is corrected.

sudarshan12s:
The fix for CI is added in this [PR](https://github.com/oracle/langchain-oracle/pull/302)

fede-kamel (approved):
Re-reviewed the full branch since my last pass, including the new embeddings class, the BaseChatModel refactor, tool calling, and the CI job. All of my previous comments are verifiably fixed:

- **SSE multi-event coalescing (the blocker):** the iterator now buffers across network chunks and splits on blank-line delimiters, so `data: {...}\n\ndata: {...}\n\n` in one read parses as two events. Covered by dedicated tests (multiple events per chunk, event split across chunks, JSON+delimiter fragmentation).
- **UTF-8 split across chunks:** `decode(..., { stream: true })` plus a final flush; test covers a split multi-byte character.
- **CR / CRLF / mixed framing:** handled, including a CRLF delimiter split across chunks.
- **stream()/invoke() text parity:** `_getChunkDataText` now joins with `""`; there's an explicit parity test.
- **Cohere history split:** now throws unless the final message is human — no more silent reordering or empty `message`.
- `_llmType` → `oci_genai`, `author` → Oracle. ✅

I ran the unit suite locally on the branch head: 89/89 pass, no type errors. The 5 failing checks on this PR are the pre-existing `make lint` breakage on main (mypy var-annotated errors from #242), fixed by #302 which is now merged — updating the branch should turn everything green.

Non-blocking notes, fine as follow-ups (some already tracked in TODO.md):

1. **`bindTools` + invoke-time `requestParams`:** `RunnableBinding` merges kwargs shallowly, so calling `boundModel.invoke(msgs, { requestParams: { temperature: 0.2 } })` replaces the whole `requestParams` object and silently drops the bound `tools`/`toolChoice`. Worth either a deep-merge of `requestParams` in `_createRequest` or a doc note.
2. **`_contentToText` rejects `content: []`:** an AIMessage with an empty content array (tool-call-only turns produced by some providers/agent frameworks) throws "Unsupported message content". Consider treating an empty array as `""`.
3. **Streaming tool-call chunk `index`** uses the per-delta array position — fine for the common cases and the code comment documents the trade-off vs Python's remapping; just flagging it as a known limitation if OCI ever streams parallel calls one-at-a-time.
4. **Embeddings `inputType`** is fixed per instance, so one instance can't send SEARCH_DOCUMENT for `embedDocuments` and SEARCH_QUERY for `embedQuery`. Matches Python parity today; noted for the Embed v4 follow-up.

Nice work on the rewrite — the SSE iterator and the lifecycle handling (single-flight init, close-during-init) are genuinely solid now.

fede-kamel:
Follow-up to my approval: also verified live against OCI GenAI (us-chicago-1, on-demand serving, session-token auth via the package's `Session` auth type). All six probes passed through the real service:

- `OciGenAiGenericChat.invoke` (meta.llama-3.3-70b-instruct): text, usage tokens, and finish_reason all populated
- `stream()`: multiple chunks over real SSE with emoji/accented text intact end-to-end (no U+FFFD), confirming the rewritten iterator handles real network chunking and streaming UTF-8 decode
- `bindTools` tool-call roundtrip: service-generated tool-call ID → `ToolMessage` → final answer
- streamed tool-call chunks reassemble correctly via `AIMessageChunk.concat`
- `OciGenAiCohereChat` invoke + stream (cohere.command-r-plus-08-2024)
- `OciGenAiEmbeddings.embedDocuments`/`embedQuery` (cohere.embed-english-v3.0), dimensions consistent

One harmless observation: the OCI SDK prints "No DEFAULT profile was specified in the configuration" to stdout when the config file has no `[DEFAULT]` section — SDK noise, not a package bug.

PR #287, opened 2026-08-07
[oracle/langchain-js]: Remove uuid dependency as it is unused

sharadraju:
[oracle/langchain-js]: Update uuid dependency version to 11.0.0 the supported version

### DIFF
--- a/libs/js/langchain-oracledb/package.json
+++ b/libs/js/langchain-oracledb/package.json
@@ -39,7 +39,7 @@
   "license": "MIT",
   "dependencies": {
     "htmlparser2": "^10.0.0",
-    "uuid": "^10.0.0"
+    "uuid": "^11.0.0"
   },
   "peerDependencies": {
     "@langchain/core": "^1.0.0",
--- a/libs/js/langchain-oracledb/package.json
+++ b/libs/js/langchain-oracledb/package.json
@@ -38,7 +38,7 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "htmlparser2": "^10.0.0"
+    "htmlparser2": "^10.0.0",
   },
   "peerDependencies": {
     "@langchain/core": "^1.0.0",
--- a/libs/js/langchain-oracledb/package.json
+++ b/libs/js/langchain-oracledb/package.json
@@ -38,8 +38,7 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "htmlparser2": "^10.0.0",
-    "uuid": "^11.0.0"
+    "htmlparser2": "^10.0.0"
   },
   "peerDependencies": {
     "@langchain/core": "^1.0.0",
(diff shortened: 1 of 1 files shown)


libs/js/langchain-oracledb/package.json
```diff
@@ -39,7 +39,7 @@
   "license": "MIT",
   "dependencies": {
     "htmlparser2": "^10.0.0",
-    "uuid": "^10.0.0"
+    "uuid": "^11.0.0"
```
sudarshan12s: we can remove it , we no longer depend on it. we used `crypto.randomUUID();`

fede-kamel (commented):
Heads-up before merging: `uuid` is declared in `libs/js/langchain-oracledb/package.json` but never imported anywhere in the package source (`grep -rn uuid src/` comes back empty — only `htmlparser2` is actually used). Rather than bumping it to ^11, I'd suggest removing the dependency entirely. Happy to approve either way if there's a use for it I'm missing (e.g. planned work that needs it).

sharadraju:
Thanks @fede-kamel . I have removed it now based on @sudarshan12s 's suggestion.

sudarshan12s (approved):

fede-kamel:
Opened #288 as the removal alternative discussed above — `uuid` is never imported in the package, so dropping the dependency resolves any version concern outright. Happy to close #288 in favor of this bump if there's planned work that needs uuid.

fede-kamel (changes requested):
Thanks for switching to removal! One catch though: the edit leaves a trailing comma behind — `"htmlparser2": "^10.0.0",` followed by `}` — which makes `package.json` invalid JSON, so `npm`/`pnpm install` will fail to parse it. CI is green only because the repo CI doesn't build the JS package. One-character fix: drop that comma. (Alternatively #288 is the same change with the comma handled, verified with `pnpm install` + `pnpm build` + unit tests — whichever you prefer, happy to close mine.)

libs/js/langchain-oracledb/package.json
```diff
@@ -39,7 +39,6 @@
   "license": "MIT",
   "dependencies": {
     "htmlparser2": "^10.0.0",
```
fede-kamel:
```suggestion
    "htmlparser2": "^10.0.0"
```
Removes the trailing comma so the JSON stays valid.

fede-kamel (commented):
Adding the fix as a one-click suggestion so this can move quickly:

sharadraju:
Thanks @fede-kamel . I have applied your suggestion here. Please merge this and close the other MR.

fede-kamel (approved):
Verified locally end-to-end: applied suggestion fixes the trailing comma, package.json parses, and no source file references uuid. Ran the full vitest suite against a local Oracle Free (full image) container with an in-database ONNX embedding model: 66 passed, 2 skipped (the DEMO_CREDENTIAL-gated tests), including vectorstores, document loaders, embeddings, summary, and end-to-end. Build, tree-shaking, and type checks also pass. Thanks @sharadraju!

PR #289, opened 2026-08-11
docs(oci): exhaustive deepagents documentation set

fede-kamel:
## Summary

Expands the deepagents documentation from a single module README into an exhaustive, structured guide set under `libs/oci/docs/deepagents/`, covering the entire `create_deepagents_agent` surface. Docs-only apart from two docstring clarifications; no behavior changes.

### New guide set

| Page | Covers |
|---|---|
| `docs/deepagents/guide.md` | Architecture, the two build paths (lightweight vs deep) and the exact decision rules, construction lifecycle (incl. factory-time network I/O), complete parameter reference, environment variables |
| `docs/deepagents/datastores.md` | `VectorDataStore` contract, ADB and OpenSearch adapter behavior contracts (chunk round-trip, `metadata.id` index, update semantics, normalization rules), custom stores, generated tools, query routing, hybrid search / RRF, embeddings |
| `docs/deepagents/persistence.md` | `OracleSaver`/`AsyncOracleSaver` checkpoints, `OracleStore`, durable virtual filesystem via `StoreBackend` — previously undocumented despite `langgraph-oracledb` being shipped by the `[deepagents]` extra for exactly this |
| `docs/deepagents/capabilities.md` | Built-in tools, subagents, skills, memory, human-in-the-loop (`interrupt_on`), filesystem permissions, structured output, middleware ordering; documents that upstream `state_schema` is not exposed |
| `docs/deepagents/operations.md` | BYO model, async usage & cleanup (`_oci_llm` / `aclose`), logging, sharp edges, troubleshooting matrix, testing |

Pages include mermaid architecture/sequence/class diagrams (rendered natively by GitHub).

### Other changes

- `langchain_oci/agents/deepagents/README.md` rewritten as overview + quickstart + doc index (content moved into the guide set, nothing dropped)
- New runnable sample `samples/11-deepagents/persistence_oracle_example.py` (OracleSaver thread resume + checkpoint listing) and samples README section for it
- Docstring clarifications in `agent.py`: `interrupt_before`/`interrupt_after` are lightweight-path only, `interrupt_on` is deep-path only, and a note that unexposed upstream params (e.g. `state_schema`) are not forwarded

### Verification

- Every behavioral claim was verified against source: `agent.py`, `agents/common.py`, `datastores/` (adapters, tools, selector, factory), `langgraph-oracledb`, and the upstream `deepagents` 0.7.5 wheel (pin is `>=0.6.1`)
[...]

PR #291, opened 2026-08-11
fix(oci): Gemini rejects tool schemas with exclusiveMinimum/exclusiveMaximum (breaks deepagents>=0.7 default path)

fede-kamel:
## Problem

Google's function-declaration API rejects any chat request whose tool parameter schemas contain `exclusiveMinimum` / `exclusiveMaximum`:

```
400 — Invalid JSON payload received. Unknown name "exclusiveMinimum" at 'tools[0].function_declarations[6].parameters.properties[4].value'
```

Pydantic v2 emits exactly that keyword for `gt=` / `lt=` field constraints, so **any tool with such a constraint breaks every `google.*` model on OCI GenAI**. The highest-impact case: `deepagents>=0.7` ships a built-in `grep` tool with `max_count: Optional[int] = Field(gt=0)`, which makes `create_deepagents_agent(...)` — whose default model is Gemini — fail on its very first model call. Discovered during end-to-end verification of the deepagents documentation (#289).

## Fix

`GeminiProvider.convert_to_oci_tool` now post-processes the converted tool definition:

- numeric (draft 2020-12) `exclusiveMinimum`/`exclusiveMaximum` → the inclusive `minimum`/`maximum` bounds Gemini accepts
- draft-4 boolean flag form is dropped (its paired inclusive bound stands on its own)
- an explicit inclusive bound always wins over a converted one
- `properties`/`$defs`/`definitions` keys are exempt from keyword matching, so a user field literally named `exclusiveMinimum` survives

Scoped to `GeminiProvider` only: Meta/OpenAI/xAI models on OCI accept these keywords (verified live), and `test_12_exclusive_native` (issue #103) intentionally asserts they're preserved for `GenericProvider` — that behavior is unchanged and now covered by an explicit regression test.

## Verification

- **E2E before/after** against OCI GenAI (us-chicago-1): the failing deep-agent call on `google.gemini-2.5-flash` (deepagents 0.7.5) now completes — built-in `write_file` runs and the file lands in graph state.
- **Unit:** 5 new tests (rewrite, the exact deepagents `grep` shape, draft-4/explicit-bound precedence, user-field-named-keyword survival, GenericProvider preservation). Full `tests/unit_tests/chat_models/` suite: **325 passed, 7 skipped**.

## Related

- #289 documents this as a known issue with workarounds; once this merges, that note can be softened to "fixed in <next release>".

### DIFF
--- a/libs/oci/langchain_oci/chat_models/providers/generic.py
+++ b/libs/oci/langchain_oci/chat_models/providers/generic.py
@@ -1099,13 +1099,76 @@ class OpenAIProvider(GenericProvider):
         return result
 
 
+def _to_gemini_compatible_schema(schema: Any) -> Any:
+    """Rewrite JSON-Schema numeric exclusive bounds for Gemini.
+
+    Google's function-declaration schema dialect rejects any request whose
+    tool parameters contain ``exclusiveMinimum`` / ``exclusiveMaximum``
+    (HTTP 400 ``Unknown name "exclusiveMinimum"``). Pydantic v2 emits the
+    numeric draft 2020-12 form for ``gt``/``lt`` constraints, so convert it
+    to the inclusive ``minimum`` / ``maximum`` bounds Gemini does accept.
+    The draft-4 boolean form is dropped (its paired inclusive bound already
+    stands on its own), and an explicit inclusive bound always wins over a
+    converted one.
+    """
+    if isinstance(schema, list):
+        return [_to_gemini_compatible_schema(item) for item in schema]
+    if not isinstance(schema, dict):
+        return schema
+
+    converted: Dict[str, Any] = {}
+    for key, value in schema.items():
+        if key in ("exclusiveMinimum", "exclusiveMaximum"):
+            # bool is the draft-4 flag form; drop it. (bool before number:
+            # isinstance(True, int) is True.)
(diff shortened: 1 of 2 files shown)


fileames:
Thanks for the fix.
Two things I’d want to point, not necessarily blocking:

1. In _to_gemini_compatible_schema, when both an explicit inclusive bound and an exclusive bound are present (e.g. {"maximum": 9, "exclusiveMaximum": 3}), the exclusive one is just dropped rather than reconciled — so the tighter constraint silently loses to the looser one. test_gemini_explicit_inclusive_bound_wins documents this as intended, but is that actually reachable from real Pydantic output, or only from hand-built schemas? If it’s only synthetic, maybe worth a comment noting why; if it’s reachable, we might want to take min(minimum, exclusiveMinimum) / max(maximum, exclusiveMaximum) instead of just discarding.

2. convert_to_oci_tool only rewrites when isinstance(parameters, dict) — if the OCI SDK ever returns something dict-like but not a plain dict, this becomes a silent no-op with no warning. Might be worth a debug log (or an assert in tests pinning the SDK’s return type) so a future SDK change doesn’t quietly reintroduce the bug this PR fixes.

Neither is a big deal given the current test coverage and E2E verification.

fileames (approved):

PR #293, opened 2026-08-11
feat(langgraph-oracledb): containment matching for dict and list checkpoint metadata filters

### LINKED ISSUE #290: feat(langgraph-oracledb): nested-path containment matching for checkpoint metadata filters in _search_where
Follow-up from the discussion in #280 (raised by @fileames).

`_search_where()` in `libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/base.py` currently uses `JSON_EQUAL(JSON_QUERY(metadata, '$.key'), :param)` for dict/list filter values, which requires an **exact match** of the whole nested object. The LangGraph.js Oracle implementation instead performs **recursive containment**: `{"nested": {"b": {"c": true}}}` should match metadata `{"nested": {"a": 1, "b": {"c": true}}}`.

Proposed fix: flatten dict filter values into nested-path predicates (the same approach #280 applies to OracleVS metadata filters), e.g. `JSON_VALUE(metadata, '$.nested.b.c') = 'true'`, recursing into nested dicts and keeping `JSON_EQUAL` only for lists (positional semantics). This also aligns Python behavior with the JS saver and with `PostgresSaver`'s containment semantics (`metadata @> filter`).

Scope note: kept separate from #280 because it changes `langgraph-oracledb` (separate package/versioning) and checkpoint-metadata filter semantics rather than vector-store filter operators.

### LINKED ISSUE #296: langgraph-oracledb: list metadata-filter semantics diverge from LangGraph.js Oracle saver and PostgresSaver (positional vs containment)
## Summary

`langgraph-oracledb`'s checkpoint metadata filters treat **list values as exact positional matches** (`JSON_EQUAL(JSON_QUERY(metadata, '$.path'), :bind)` with `json.dumps(value)`), while the LangGraph.js Oracle saver (langchain-ai/langgraphjs#2660) and `PostgresSaver` (`metadata @> filter`) both use **order-insensitive containment**. The two Oracle packages share the same tables and advertise cross-language compatibility, so the same `list(filter=...)` call returns different rows depending on which language issues it.

## Reproduction (Oracle Database 23ai Free, same checkpoint row)

Checkpoint metadata: `{"tags": ["a", "b"]}` — queried via `saver.list(None, filter=...)` from each language:

| filter | Python (`langgraph-oracledb`) | JS (`@langchain/langgraph-checkpoint-oracledb`) |
|---|---|---|
| `{"tags": ["a", "b"]}` | 1 match | 1 match |
| `{"tags": ["b", "a"]}` | 0 matches | 1 match |
| `{"tags": ["a"]}` | 0 matches | 1 match |

## Where

`langgraph_oracledb/checkpoint/oracle/base.py`, `_search_where` → `_add_leaf_condition`, the `isinstance(value, list)` branch:

```python
elif isinstance(value, list):
    # Lists keep exact-match semantics (element order matters)
    param_name = _next_param()
    filter_conditions.append(
        f"JSON_EQUAL(JSON_QUERY(metadata, '$.{path}'), :{param_name})"
    )
    param_values[param_name] = json.dumps(value)
```

Note both implementations' comments currently claim to match the other: this branch's docstring says the dict handling "match[es] the LangGraph.js Oracle saver and PostgresSaver", while the JS compiler's array branch is documented as Python-compatible. Dicts do agree (recursive containment, #290); lists don't.

## Proposal

Align lists to containment, matching JS and Postgres: the stored value at the path must be an array, and every filter element must appear somewhere in it (order-insensitive, duplicates not counted). The JS compiler's shape translates directly to Oracle JSON path predicates:

- `JSON_EXISTS(metadata, '$.path?(@.type() == "array")')`
- per element: `JSON_EXISTS(metadata, '$.path[*]?(@.type() == "string" && @ == $VAR)' PASSING :bind AS "VAR")` (typed predicates per element type; nested dict/array elements compile recursively, as in JS `compileMetadataFilterValue`)

This is a behavior change for callers relying on positional equality, but it restores cross-language consistency and matches the Postgres reference semantics.

Related: #290 (dict containment, PR #293), langchain-ai/langgraphjs#2660 (JS Oracle saver review where the divergence was found).

fede-kamel:
## Summary

Closes #290 — the follow-up @fileames raised on #280.
Closes #296 — list-filter divergence from the LangGraph.js Oracle saver, found while reviewing langchain-ai/langgraphjs#2660.

Metadata filters in `list()`/`alist()` previously required **exact** `JSON_EQUAL` matches: dict filters didn't match metadata containing extra keys, and list filters required exact positional equality. Both diverged from the LangGraph.js Oracle saver and from `PostgresSaver`'s containment semantics (`metadata @> filter`).

### Dict containment (#290)

`_search_where()` flattens dict filter values into per-leaf JSON-path predicates — the same nested-path approach #280 applied to OracleVS:

```sql
JSON_VALUE(metadata, '$.nested.b.c') = 'true'
```

- recurses through arbitrarily deep dicts; **every nested key** is validated with the same JSON-path-injection check as top-level keys
- leaf typing preserved: `RETURNING NUMBER` for numerics, literal `'true'`/`'false'` for booleans, `IS NULL` for nulls, bind params for strings
- `{}` asserts path existence via `JSON_EXISTS`
- sync (`OracleSaver`) and async (`AsyncOracleSaver`) share the implementation

### List containment (#296)

List filter values now match any stored array that **contains every filter element**, regardless of order or duplicates — the semantics the JS saver and `PostgresSaver` already use. The stored value must be an array (scalars never match), and elements compare by JSON type:

```sql
JSON_EXISTS(metadata, '$.tags?(@.type() == "array")')
JSON_EXISTS(metadata, '$.tags[*]?(@.type() == "string" && @ == $FILTER_KEY_0)'
            PASSING :filter_key_0 AS "FILTER_KEY_0")
```

- string/number elements are bound through the `PASSING` clause; only our own literals (`true`, `false`, `null`, `""`) reach the path text
- dict elements match by recursive containment, array elements by nested containment; nested member names go through the same injection validation
- `[]` matches any stored array at the path
- verified live against the JS saver on shared tables: for metadata `{"tags": ["a","b"]}`, the filters `["a","b"]` / `["b","a"]` / `["a"]` now return identical results from both languages (previously Python returned 1/0/0 vs JS 1/1/1), and a seven-case typed-element matrix (dict containment, dict miss, `null`, `true`/`false` discrimination, numbers, absent nested arrays) agrees exactly

Backward compatibility: an exact-match filter is trivially contained in equal metadata, so every previously-matching filter still matches — the change only widens matching, it never narrows it.

## Tests

- `tests/test_search_where_nested_containment.py`: SQL-generation assertions (no DB) + live sync/async containment tests — dict scenarios from #290 plus list containment (order-insensitivity, subsets, non-matching elements, empty-list, typed elements incl. dict-in-list), injection checks for nested dict keys and dict-in-list member names, non-finite number rejection
- `test_search_where_key_combinations.py` / `test_search_where_parameter_combinations.py` updated to the containment contract; previously-disabled complex list cases (`[1, 2, "three"]`, `[{"item": 1}, {"item": 2}, [3, 4]]`) re-enabled
- Full package suite verified against a local Oracle 26ai Free container (per-file, see note), all green

Note: this package has no repo CI job, so the verification above was run locally. On a resource-constrained container (default `processes=200`) the suite must be run file-by-file or with pauses — running all files back-to-back exhausts DB processes and fails spuriously with `ORA-12516`.

### DIFF
--- a/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/base.py
+++ b/libs/langgraph-oracledb/langgraph_oracledb/checkpoint/oracle/base.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import math
 import random
 from collections.abc import Sequence
 from decimal import Decimal
@@ -462,7 +463,9 @@ class BaseOracleSaver(BaseCheckpointSaver[str]):
             # metadata {"nested": {"a": 1, "b": {"c": True}}}. Each dict is
             # flattened into per-leaf JSON-path predicates, matching the
             # LangGraph.js Oracle saver and PostgresSaver (`metadata @> filter`).
-            # Lists keep JSON_EQUAL exact-match (positional) semantics.
+            # Lists also use containment (issue #296): the stored value must be
+            # an array containing every filter element, regardless of order or
+            # duplicates, with dict/array elements matched recursively.
             filter_conditions = []
             param_counter = 0
 
@@ -471,6 +474,72 @@ class BaseOracleSaver(BaseCheckpointSaver[str]):
                 name = f"filter_key_{param_counter}"
                 param_counter += 1
                 return name
+
+            def _bound_path_variable(value: Any, passing: list[str]) -> str:
+                """Bind a value for use as a JSON path variable ($NAME)."""
+                param_name = _next_param()
+                variable = param_name.upper()
+                passing.append(f':{param_name} AS "{variable}"')
+                param_values[param_name] = value
+                return f"${variable}"
+
+            def _is_object_predicate(subject: str) -> str:
+                return " && ".join(
+                    f'!({subject}.type() == "{scalar}")'
+                    for scalar in ("array", "string", "number", "boolean", "null")
+                )
+
+            def _element_predicate(subject: str, value: Any, passing: list[str]) -> str:
+                """JSON-path predicate matching one array element by containment.
+
+                Mirrors the LangGraph.js Oracle saver's filter compiler so both
+                languages match the same rows: scalars compare by type and
+                value, dicts by recursive containment, arrays by nested
+                containment. String and numeric values are bound through the
+                PASSING clause; only our own literals reach the path text.
+                """
+                if value is None:
+                    return f'{subject}.type() == "null" && {subject} == null'
+                if isinstance(value, bool):
+                    literal = "true" if value else "false"
+                    return f'{subject}.type() == "boolean" && {subject} == {literal}'
+                if isinstance(value, (int, float)):
+                    if isinstance(value, float) and not math.isfinite(value):
+                        raise ValueError(
+                            "Metadata filter numbers inside lists must be finite."
+                        )
+                    variable = _bound_path_variable(value, passing)
+                    return f'{subject}.type() == "number" && {subject} == {variable}'
+                if isinstance(value, str):
+                    if value == "":
+                        return f'{subject}.type() == "string" && {subject} == ""'
+                    variable = _bound_path_variable(value, passing)
+                    return f'{subject}.type() == "string" && {subject} == {variable}'
+                if isinstance(value, list):
+                    predicates = [f'{subject}.type() == "array"']
+                    predicates.extend(
+                        f"exists({subject}[*]?"
+                        f"({_element_predicate('@', item, passing)}))"
+                        for item in value
+                    )
+                    return " && ".join(predicates)
+                if isinstance(value, dict):
+                    predicates = [] if subject == "@" else [f"exists({subject})"]
+                    predicates.append(_is_object_predicate(subject))
+                    for sub_key, sub_value in value.items():
+                        # SECURITY: member names become part of the JSON path
+                        # expression; validate exactly like top-level keys.
+                        self._validate_json_path_key(sub_key)
+                        predicates.append(
+                            _element_predicate(
+                                f'{subject}."{sub_key}"', sub_value, passing
+                            )
+                        )
+                    return " && ".join(predicates)
+                raise ValueError(
+                    "Metadata filter list elements must be JSON values, got "
+                    f"{type(value).__name__}."
+                )
 
             def _add_leaf_condition(path: str, value: Any) -> None:
                 """Append a predicate comparing one JSON path to a leaf value."""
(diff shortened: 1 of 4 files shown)


fileames:
Thanks a lot for working on this.
Went through this against the JS saver since that's what it's aligning to. The containment approach looks right, and `_element_predicate` lines up well with how the JS filter compiler does it (type-checked scalars, values bound through PASSING, recursive dicts and arrays).

Two things I want to consider.

`_expand_filter` builds unquoted paths ('$.{path}') while `_element_predicate` quotes member names ({subject}."{sub_key}"). So a filter key containing a dot is a nested path at the top level but a literal key inside a list element. Before this PR only top-level keys got interpolated into a path (dicts went through JSON_EQUAL on the whole subdocument), so the flattening spreads that ambiguity to every depth. JS quotes everywhere and reads {"a.b": 1} as a literal key. Can we quote the flattened segments too, or is dotted-as-nested intentional? 

Smaller one: `_element_predicate` rejects non-finite floats, but the numeric branch in `_add_leaf_condition` doesn't, so {"score": float("nan")} still gets bound.

One more thing we can fix, though it is not directly same as the main issue. The boolean leaf compares `JSON_VALUE(...) = 'true'`, so it also matches the string "true"; IS NULL matches an absent path as well as an explicit JSON null; and {"a": {}} only checks existence, so it matches {"a": 5}. In each of those JS agrees with what metadata @> filter does in Postgres, so I'd say Python is the odd one out and should follow.

The catch is that unlike containment these would narrow matching rather than widen it, so they'd change results for filters people are relying on today ({"x": None} currently finds rows that don't have the key at all). Do you think we should make the change or a new PR with clear release note?

fede-kamel:
Thanks @fileames, both points reproduced live and addressed in the latest commit.

**Dotted keys.** You were right that it was ambiguous. The key validator accepts dots, so `{"a.b": 1}` at the top level or inside a nested dict compiled to the path `$.a.b`, while the same key inside a list element compiled to the literal member `@."a.b"`. I went with literal keys everywhere: every path segment is now rendered as a quoted member name (`$."user"."a.b"`), at every depth, which matches the JS saver and PostgresSaver. Verified on 26ai: a checkpoint with metadata `{"a.b": 1, "a": {"b": 2}}` is matched by `{"a.b": 1}` and by `{"a": {"b": 2}}`, and not by `{"a": {"b": 1}}`. For plain keys the generated SQL is equivalent to before. Dotted-as-nested was never documented, so I don't think anyone depends on it, but it is a behaviour change for anyone who discovered it and I've called it out in the commit message.

**NaN.** Confirmed: `{"score": float("nan")}` at a top-level or nested numeric leaf was accepted in Python and failed in the driver with `DPY-4004: invalid number`. It now raises `ValueError("Metadata filter numbers must be finite.")`, same as the list-element guard.

**Narrowing semantics** (bool compared as the string `'true'`, `IS NULL` matching absent keys, `{}` matching scalars): agreed that Python is the odd one out. Since those changes narrow existing matches rather than widen them, I'd keep them out of this PR and do them together in a follow-up with a release note, so this one stays a pure containment/widening change. Happy to open that PR right after this merges.

Verification on this commit: full langgraph-oracledb suite 861 passed / 122 skipped against Oracle AI Database 26ai Free (main is 809 / 122 on the same DB, so +52 tests, no regressions), plus a live probe over nested-dict, list-containment, empty-list, empty-dict, mixed scalar and injection-key cases.

PR #292, opened 2026-08-11
chore: bump langchain-oracledb to 1.6.0

fede-kamel:
Version bump for release:

* **langchain-oracledb 1.5.0 → 1.6.0**: exact-match OracleCache (#282), OracleByteStore/OracleDocStore (#283), nested metadata filters in OracleVS (#280)

Note: this PR originally also bumped langgraph-oracledb to 1.0.2, but that bump landed on main via #295 (which pairs it with the Decimal index-param fix for #294), so it was dropped here on rebase. If #293 (containment matching for checkpoint metadata filters) merges before the langgraph-oracledb release is cut, consider bumping to 1.1.0 in that context since it changes filter-matching behavior.

PR #295, opened 2026-08-12
fix(langgraph-oracledb): coerce Decimal index params before comparison

### LINKED ISSUE #294: bug(oracledb): OracleStore.setup() fails with TypeError: Object of type Decimal is not JSON serializable when an index config has a numeric index_type parameter
Package: langgraph-oracledb 1.0.1 — Affects: OracleStore and AsyncOracleStore

Summary

Calling setup() a second time on a store created with an explicit table_suffix and an index_type containing any numeric parameter raises a TypeError. The store becomes unusable from that point on: the tables can be created once and then never reopened.

Reproduction
```

from langgraph_oracledb.store.oracle import OracleStore

CONN = "user/password@localhost:1521/FREEPDB1"
index = {
    "dims": 8,
    "embed": my_embeddings,
    "fields": ["text"],
    "index_type": {"type": "ivf", "neighbor_partitions": 1},   # any number here
}

with OracleStore.from_conn_string(CONN, index=index, table_suffix="demo") as store:
    store.setup()          # ok — registers the configuration

with OracleStore.from_conn_string(CONN, index=index, table_suffix="demo") as store:
    store.setup()          # TypeError
```

Actual result

  File ".../store/oracle/base.py", line 1325, in setup
    self._validate_configuration(self.table_suffix, self.index_config)
  File ".../store/oracle/base.py", line 1177, in _validate_configuration
    existing_params_normalized = json.dumps(existing_params_dict, sort_keys=True)
TypeError: Object of type Decimal is not JSON serializable

Cause

STORE_CONFIGS.index_params is an Oracle JSON column, and python-oracledb returns its numbers as decimal.Decimal. _normalize_existing_index_params (base.py:352) returns the mapping unchanged, and _validate_configuration hands it straight to json.dumps (base.py:1182, :1193). AsyncOracleStore shares the defect through the same helper (aio.py:236, :247).

The checkpoint side already solves this — BaseOracleSaver._coerce_decimals (checkpoint/oracle/base.py:603) exists because "Oracle's native JSON type deserializes every number as Decimal" — but the store never applies it.

Trigger conditions

All three required: an explicit table_suffix (a derived suffix sets _needs_validation = False); at least one numeric index_type value (neighbors, efconstruction, neighbor_partitions, samples_per_partition, min_vectors_per_partition); and a configuration already registered, i.e. any run after the first. An index_type with no numeric parameter is unaffected — which is why default configurations never hit it. accuracy alone is also safe, since it's popped before the dump.

fileames:
OracleStore.setup() raised "TypeError: Object of type Decimal is not JSON serializable" whenever it re-validated a registered index configuration that contained a numeric index_type parameter, making the store unusable after the first run.

STORE_CONFIGS.index_params is an Oracle JSON column, so python-oracledb returns its numbers as Decimal. _validate_configuration compares configurations by json.dumps()-ing the normalized dict, and json has no encoder for Decimal.

_normalize_existing_index_params now coerces Decimal to int/float, which fixes both OracleStore and AsyncOracleStore since they share the helper. The same coercion already exists on the checkpoint side (BaseOracleSaver._coerce_decimals) for exactly this reason.

Triggered by an explicit table_suffix plus any numeric index_type value (neighbors, efconstruction, neighbor_partitions, samples_per_partition, min_vectors_per_partition) on the second setup(). Configurations without a numeric index_type parameter were unaffected, which is why defaults never hit it.

Fixes #294

### DIFF
--- a/libs/langgraph-oracledb/langgraph_oracledb/store/oracle/base.py
+++ b/libs/langgraph-oracledb/langgraph_oracledb/store/oracle/base.py
@@ -11,6 +11,7 @@ import threading
 from collections import defaultdict
 from collections.abc import Iterable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
+from decimal import Decimal
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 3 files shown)


fede-kamel (approved):
Verified end-to-end. Reproduced the bug on the released 1.0.1 against a local Oracle 23ai Free container: an OracleStore with a numeric ivf index param (neighbor_partitions) fails on the second setup() with 'TypeError: Object of type Decimal is not JSON serializable', exactly as #294 describes. On this branch the same repro passes, and tests/test_store_config_isolation.py (26), test_store_setup_schema_consistency.py + test_store_search_sync.py (13) are all green against the live DB. The fix mirrors BaseOracleSaver._coerce_decimals as documented, and covers both the Mapping and JSON-string paths. One heads-up: the 1.0.2 version bump here overlaps with #292 — I'll rebase that one after this merges.

PR #298, opened 2026-08-13
fix(js/langchain-oracledb): preserve sparse vector element format (FLOAT32/INT8) to prevent ORA-51812

### LINKED ISSUE #297: SPARSE FLOAT32 similarity search fails with ORA-51812 because query vector is encoded as FLOAT64
### Problem

  When `OracleVS` is configured with `vectorType: SPARSE` and `format: FLOAT32`, similarity search fails because the sparse query vector    is encoded as `FLOAT64` instead of `FLOAT32`.

  ### Environment

  - `@oracle/langchain-oracledb`: 1.0.0
  - `oracledb`: 7.0.1
  - Node.js: 24.18.0
  - Oracle Database: 23.26.2.0
  - node-oracledb mode: Thin

  ### Reproduction

  ```js
  const store = new OracleVS(embeddings, {
    client: connection,
    tableName: "SPARSE_TEST",
    query: "dimension probe",
    vectorType: VectorType.SPARSE,
    format: VectorElementFormat.FLOAT32,
  });

  await store.initialize();

  await store.addVectors(
    [[1, 0, 0, 0.5, 0]],
    [new Document({ pageContent: "test" })]
  );

  await store.similaritySearchByVectorReturningEmbeddings(
    [1, 0, 0, 0.5, 0],
    1
  );
```
  ### Actual behavior

  The similarity search fails with: `ORA-51812: The vector SQL function VECTOR_DISTANCE() requires all input vectors
  to have the same dimension format: encountered dimension formats
  (FLOAT32, FLOAT64)`

  The table is created with the expected column definition: `VECTOR(5, FLOAT32, SPARSE)`

  However, `prepareVectorForStorage()` constructs the sparse vector using the dense-input form of the node-oracledb `SparseVector` constructor: `new oracledb.SparseVector(new Float32Array(vector))`

  Although the input is explicitly a Float32Array, this constructor path produces a sparse vector whose nonzero values are stored in a  `Float64Array`:

  ```js
  const input = new Float32Array([1, 0, 2, 0]);
  const sparseVector = new oracledb.SparseVector(input);

  console.log(sparseVector.values.constructor.name);
  // Float64Array
```
  Consequently, the query vector is encoded as FLOAT64. Oracle then compares it with the FLOAT32 sparse vector stored in the column, causing VECTOR_DISTANCE() to raise ORA-51812.

  ### Expected behavior

  ### Expected behavior

  When `OracleVS` is configured with:

  ```js
  vectorType: VectorType.SPARSE,
  format: VectorElementFormat.FLOAT32
```
  both inserted and query vectors should be bound as sparse FLOAT32 vectors so that VECTOR_DISTANCE() can execute successfully.

  If a FLOAT64 vector is supplied, convertDenseVectorForFormat() should convert it to the configured vectorFormat. This conversion   currently works and produces a Float32Array. However, passing that result to new oracledb.SparseVector() converts its nonzero values back to a Float64Array, effectively undoing the format conversion.

### PR DESCRIPTION
**Pull request by fede-kamel**
Fixes #297

## Problem

With `vectorType: SPARSE` and `format: FLOAT32`, similarity search fails with `ORA-51812: ... encountered dimension formats (FLOAT32, FLOAT64)`. Root cause (verified against node-oracledb 7.0.1): the dense-input `oracledb.SparseVector` constructor stores its `values` as a `Float64Array` **regardless of the typed array passed in**, so the query vector reaches `VECTOR_DISTANCE()` as FLOAT64 while the column is FLOAT32:

```js
new oracledb.SparseVector(new Float32Array([1, 0, 2, 0.5])).values.constructor.name  // "Float64Array"
new oracledb.SparseVector(new Int8Array([1, 0, 2, 0])).values.constructor.name       // "Float64Array" — INT8 sparse was equally broken
```

## Fix

`prepareVectorForStorage()` now extracts the nonzero `indices`/`values` into the correctly-typed array and uses the object-form constructor, which preserves the element format:

```js
new oracledb.SparseVector({ values: new Float32Array(...), indices, numDimensions })
```

Semantics preserved from the previous dense construction: INT8 values are rounded and range-checked before zero-dropping (a value rounding to 0 is dropped, as the dense path produced), and FLOAT32 uses `Math.fround` for the zero test so values that underflow to 0 in float32 are dropped as before.

## Testing

- **Live repro** on Oracle Database 23ai Free (node-oracledb 7.0.1, Thin): before the fix, the issue's exact scenario fails with ORA-51812 on a `VECTOR(5, FLOAT32, SPARSE)` column; with the fix, both the issue's FLOAT32 scenario and the INT8 variant run `initialize → addVectors → similaritySearchByVectorReturningEmbeddings` successfully end-to-end.
- **Unit tests**: new `sparse vector element format (#297)` describe block pins the values' typed array per format (FLOAT32/FLOAT64/INT8), indices/numDimensions correctness, INT8 range validation, rounded-zero dropping, and the all-zero vector edge case. `vitest run src/tests/vectorstores.unit.test.ts`: 8 passed, type check clean.
- `pnpm lint` and `pnpm build` clean.

Note: this package has no CI coverage in this repo — the results above are from a local run.

### DIFF
--- a/libs/js/langchain-oracledb/src/vectorstores.ts
+++ b/libs/js/langchain-oracledb/src/vectorstores.ts
@@ -709,24 +709,55 @@ export class OracleVS extends VectorStore {
         throwError(ErrorCode.VECTOR_INVALID_VALUE, "Sparse vectors must supply full-dimension arrays for conversion.");
       }
 
-      let sparseInput: number[] | Float32Array | Float64Array | Int8Array = vector;
+      // The dense-input SparseVector constructor stores its values as a
+      // Float64Array regardless of the typed array passed in, so FLOAT32 and
+      // INT8 query vectors reach the database as FLOAT64 and VECTOR_DISTANCE()
+      // fails with ORA-51812 against a non-FLOAT64 sparse column. The
+      // object-form constructor preserves the values' typed array, so build
+      // the nonzero indices/values explicitly and use that form.
+      const indices: number[] = [];
+      const nonzero: number[] = [];
       if (format === VectorElementFormat.INT8) {
-        const clamped = new Int8Array(vector.length);
         for (let i = 0; i < vector.length; i += 1) {
           const rounded = Math.round(vector[i]);
           if (rounded < -128 || rounded > 127) {
             throwError(ErrorCode.VECTOR_INVALID_VALUE, "INT8 sparse vector values must be within [-128, 127].");
           }
-          clamped[i] = rounded;
+          if (rounded !== 0) {
+            indices.push(i);
+            nonzero.push(rounded);
+          }
         }
-        sparseInput = clamped;
-      } else if (format === VectorElementFormat.FLOAT64) {
-        sparseInput = new Float64Array(vector);
       } else if (format === VectorElementFormat.FLOAT32) {
-        sparseInput = new Float32Array(vector);
+        for (let i = 0; i < vector.length; i += 1) {
+          if (Math.fround(vector[i]) !== 0) {
+            indices.push(i);
+            nonzero.push(vector[i]);
+          }
+        }
+      } else {
+        for (let i = 0; i < vector.length; i += 1) {
+          if (vector[i] !== 0) {
+            indices.push(i);
+            nonzero.push(vector[i]);
+          }
+        }
       }
 
-      return new oracledb.SparseVector(sparseInput);
+      let values: Float32Array | Float64Array | Int8Array;
+      if (format === VectorElementFormat.INT8) {
+        values = new Int8Array(nonzero);
+      } else if (format === VectorElementFormat.FLOAT64) {
+        values = new Float64Array(nonzero);
+      } else {
+        values = new Float32Array(nonzero);
+      }
+
+      return new oracledb.SparseVector({
+        values,
+        indices,
+        numDimensions: vector.length,
+      });
     }
 
     if (vector.length !== dimension) {
--- a/libs/js/langchain-oracledb/src/vectorstores.ts
+++ b/libs/js/langchain-oracledb/src/vectorstores.ts
@@ -719,9 +719,13 @@ export class OracleVS extends VectorStore {
       const nonzero: number[] = [];
       if (format === VectorElementFormat.INT8) {
         for (let i = 0; i < vector.length; i += 1) {
-          const rounded = Math.round(vector[i]);
-          if (rounded < -128 || rounded > 127) {
-            throwError(ErrorCode.VECTOR_INVALID_VALUE, "INT8 sparse vector values must be within [-128, 127].");
+          const value = vector[i];
+          const rounded = Math.round(value);
+          if (!Number.isFinite(value) || rounded < -128 || rounded > 127) {
+            throwError(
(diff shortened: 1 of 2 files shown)


libs/js/langchain-oracledb/src/vectorstores.ts
```diff
@@ -709,24 +709,55 @@ export class OracleVS extends VectorStore {
...
+      // INT8 query vectors reach the database as FLOAT64 and VECTOR_DISTANCE()
+      // fails with ORA-51812 against a non-FLOAT64 sparse column. The
+      // object-form constructor preserves the values' typed array, so build
+      // the nonzero indices/values explicitly and use that form.
+      const indices: number[] = [];
+      const nonzero: number[] = [];
       if (format === VectorElementFormat.INT8) {
-        const clamped = new Int8Array(vector.length);
         for (let i = 0; i < vector.length; i += 1) {
           const rounded = Math.round(vector[i]);
```
sudarshan12s, line 722: can we also add `isFinite` to avoid `Nan`
```

for (let i = 0; i < vector.length; i += 1) {
          const value = vector[i];
          const rounded = Math.round(value);
          if (
            !Number.isFinite(value) ||
            rounded < -128 ||
            rounded > 127
          ) {
            throwError(
              ErrorCode.VECTOR_INVALID_VALUE,
              "INT8 sparse vector values must be finite and within [-128, 127].",
            );
          }
          if (rounded !== 0) {
            indices.push(i);
            nonzero.push(rounded);
          }
        }
```

libs/js/langchain-oracledb/src/vectorstores.ts
```diff
@@ -709,24 +709,55 @@ export class OracleVS extends VectorStore {
...
+        values = new Int8Array(nonzero);
+      } else if (format === VectorElementFormat.FLOAT64) {
+        values = new Float64Array(nonzero);
+      } else {
+        values = new Float32Array(nonzero);
+      }
+
+      return new oracledb.SparseVector({
+        values,
+        indices,
```
sudarshan12s, line 758: `indices: new Uint32Array(indices),
`

optional change :)  The same thing is however done in node-oracledb

sudarshan12s (approved):
Thanks for the changes.

LGTM.

minor comments.

sharadraju:
@fede-kamel Thanks for raising this PR. Can we add the minor suggestions provided by @sudarshan12s and merge this ?
 It will go in as part of the next release.

fede-kamel:
Applied both suggestions in the latest commit:

- `Number.isFinite` guard in the INT8 loop (message now reads "INT8 sparse vector values must be finite and within [-128, 127]."), with a unit test for `NaN` / `±Infinity`.
- `indices: new Uint32Array(indices)` in the `SparseVector` constructor call, with a unit test asserting the typed array.

Re-verified end to end against Oracle AI Database 26ai Free (23.26.2): the sparse/INT8/binary vectorstore integration tests pass, and the issue #297 reproduction (sparse FLOAT32 and INT8 `similaritySearch`) passes on this branch while `main` still fails both with ORA-51812. Unit tests, lint and type-check are clean.

@sharadraju ready for the next release from my side; @fileames could you approve when you get a moment?

fileames (approved):

Issue #263, opened 2026-07-22
feat(oci): native async for OCIGenAIEmbeddings, OCIModelDeploymentEndpointEmbeddings, and the OCIGenAI LLM


fede-kamel:
## Motivation

`ChatOCIGenAI` already implements real native async (`_agenerate`/`_astream` over an async OCI client, `libs/oci/langchain_oci/chat_models/async_mixin.py`), but the embeddings classes and the completion LLM still fall back to `langchain_core`'s default thread-pool wrappers. In async-first production stacks (FastAPI services, LangGraph async graphs), thread-pool fallbacks limit throughput and hide backpressure.

Peer packages (langchain-openai, langchain-aws) ship native async embeddings.

## Current state

- `OCIGenAIEmbeddings` (`libs/oci/langchain_oci/embeddings/oci_generative_ai.py`): sync `embed_documents`/`embed_query` only; `aembed_*` inherit defaults. Image embedding mixin is also sync-only.
- `OCIModelDeploymentEndpointEmbeddings` (`libs/oci/langchain_oci/embeddings/oci_data_science_model_deployment_endpoint.py`): sync only.
- `OCIGenAI` completion LLM (`libs/oci/langchain_oci/llms/oci_generative_ai.py`): sync `_call`/`_stream` only, no `_acall`/`_astream`.

## Proposed scope

- Implement `aembed_documents` / `aembed_query` (and async image-embedding variants) using the same async client infrastructure the chat mixin uses.
- Implement `_acall` / `_astream` on the `OCIGenAI` LLM.
- Preserve batching semantics (`batch_size=96`) in async paths.

## Acceptance criteria

- Async methods issue no thread-pool executor calls (true native async I/O).
- Unit tests with mocked async client; embeddings standard tests updated.

PR #300, opened 2026-08-18
feat(oci): delegate async paths to the SDK's native AsyncGenerativeAiInferenceClient

### LINKED ISSUE #299: feat(oci): adopt the OCI Python SDK's native AsyncGenerativeAiInferenceClient for async paths
## Summary

`langchain-oci` currently implements native async (`ainvoke`/`astream`/`aembed_documents`/`arerank`, #263) through a hand-rolled aiohttp client — `langchain_oci/common/async_support.py::OCIAsyncClient` — because the `oci` Python SDK historically exposed no async client. That module re-implements, in this repo:

- OCI request signing (via `signer(prepared_request)` on a synthetic `requests.Request`)
- SSE parsing for streaming chat / generate_text
- session management, timeouts, and error surfacing (`OCIAsyncRequestError`)

An upcoming `oci` Python SDK release adds a **native `oci.generative_ai_inference.AsyncGenerativeAiInferenceClient`** backed by a new aiohttp-based `AsyncBaseClient`, with:

- `async` `chat`, `embed_text`, `generate_text`, `rerank_text`, `summarize_text`, `apply_guardrails`, `list_guardrail_versions`
- SSE streaming via `AsyncBaseClient.call_api_stream` (yields parsed event dicts — same shape `OCIAsyncClient` yields today)
- the SDK's standard retry strategies, circuit breaker, and `oci.exceptions.ServiceError` error surface
- `close()` / async-context-manager lifecycle, lazy import so the sync-only footprint is unchanged

We should adopt the native client so signing/auth fixes and new auth methods come from the SDK instead of being duplicated here.

## Proposal

Keep `OCIAsyncClient` as the internal facade (its interface is consumed by 4 modules) and make it **delegate to `AsyncGenerativeAiInferenceClient` when the installed SDK provides it**, falling back to the current hand-rolled aiohttp transport otherwise:

- Feature-detect at import (`from oci.generative_ai_inference import AsyncGenerativeAiInferenceClient` → `ImportError` on older SDKs).
- Construct the native client with the same `config` + `signer` + `service_endpoint` the sync client was built with, and `skip_deserialization=True`, so responses remain the camelCase wire dicts all async parsing already consumes — no consumer changes.
- Non-streaming ops map 1:1 (`chat`, `embed_text`, `rerank_text`, `generate_text`).
- Streaming chat/generate use `async_base_client.call_api_stream(...)` (the SDK's documented SSE surface; the generated `chat()` is non-streaming only).
- Translate `oci.exceptions.ServiceError` → `OCIAsyncRequestError(status, body)` so the parameter-compat 400 retry (`common/param_compat.py`) and existing `except` contracts keep working on both paths.

Consumers that stay untouched by design:

- `chat_models/async_mixin.py` (`ChatOCIGenAI` `_agenerate`/`_astream`)
- `embeddings/oci_generative_ai.py`
- `document_compressors/oci_generative_ai_rerank.py`
- `llms/oci_generative_ai.py`

Once the minimum supported `oci` version includes the async client, the fallback transport can be deleted in a follow-up.

## Benefits

- One signing implementation (SDK-maintained) instead of a parallel one — new auth methods and signing fixes arrive for free.
- Retry-strategy and circuit-breaker parity between sync and async paths.
- Standard `ServiceError` metadata (request id, operation name, API reference link) available on async failures.

## Acceptance criteria

- [ ] All async ops (`chat` invoke/stream, `embed_text`, `rerank_text`, `generate_text` invoke/stream) delegate to the native client when available.
- [ ] Existing behavior preserved on currently released `oci` versions (fallback path; `oci>=2.173.0` floor unchanged).
- [ ] `ServiceError` → `OCIAsyncRequestError` mapping keeps the gpt-5-style 400 parameter-compat retry working on the native path.
- [ ] Unit tests cover both transports; live smoke test of invoke/stream/tools/embeddings/rerank against OCI GenAI with the new SDK.

fede-kamel:
Closes #299

## What

When the installed `oci` SDK provides the native `oci.generative_ai_inference.AsyncGenerativeAiInferenceClient` (added in an upcoming SDK release), `OCIAsyncClient` now delegates every operation to it instead of signing and sending requests itself:

- **chat** — non-streaming via the generated `chat()`; SSE streaming via `AsyncBaseClient.call_api_stream` (the SDK's streaming surface; it yields the same parsed event dicts the built-in SSE parser yields)
- **embed_text**, **rerank_text**, **generate_text** (invoke + stream)

On older SDKs (down to the current `oci>=2.173.0` floor) nothing changes — the built-in aiohttp transport keeps handling everything, so this ships safely ahead of the SDK release.

## How

- The native client is constructed with the **same `config` + `signer` + `service_endpoint` the sync client was built with**, so authentication behaves identically across sync and async, and with `skip_deserialization=True` so responses remain the camelCase wire dicts all existing async parsing consumes — zero changes in the four consumers (`chat_models/async_mixin.py`, `embeddings`, `document_compressors`, `llms`).
- `oci.exceptions.ServiceError` is translated to `OCIAsyncRequestError(status, body)` with the OCI error envelope as body, so existing `except` contracts and the parameter-compatibility 400 retry (`common/param_compat.py`) work unchanged on both transports.
- Native-client construction failures log at debug level and fall back to the built-in transport rather than breaking async support.
- Delegation is feature-detected at construction (`ImportError` on older SDKs → fallback), so no dependency changes.

## Why

The hand-rolled `OCIAsyncClient` transport (#263) duplicates OCI request signing and SSE parsing inside this repo. Delegating to the SDK means signing fixes and new auth methods arrive from the SDK, plus retry-strategy/circuit-breaker parity with the sync client and standard `ServiceError` metadata (opc-request-id, operation name) on failures.

## Testing

- **Unit**: new `tests/unit_tests/common/test_native_async_client.py` (15 tests) covers delegation for all ops (invoke + stream), constructor wiring (`signer`/`service_endpoint`/`skip_deserialization`), `ServiceError` translation (including the double-encoded gpt-5-style 400 round-tripping through `adjust_request_for_param_error`), fallback when the class is missing or construction fails, and `ChatOCIGenAI._agenerate`/`_astream` end-to-end over a fake native client.
- **Full suite**: 517 passed, 47 skipped on the released SDK (fallback path). `ruff` and `mypy` clean.
- **Live** (tested with OCI GenAI on a prerelease SDK build; verified `_native` is the SDK client, not the fallback):
  - `ainvoke` (meta.llama-3.3-70b-instruct) — correct content + usage metadata
  - `astream` — multi-chunk stream with multibyte UTF-8 intact (`café ☕ 日本語`)
  - `bind_tools` + `ainvoke` — tool call extracted with id/args
  - `aembed_documents` (cohere.embed-english-v3.0) — 1024-dim vectors
  - `openai.gpt-5` with non-default `temperature` — the 400 parameter-compat retry engages through the translated error and the call succeeds
  - Live `rerankText` returns 404 on the test tenancy (pre-existing tenancy limitation documented in `tests/integration_tests/document_compressors/test_oci_generative_ai_rerank.py`); the failure surfaced exactly per contract as `OCIAsyncRequestError(404, <envelope>)`, and delegation itself is unit-tested.


fede-kamel:
## Extended live verification — multi-model matrix over the native async client

Ran the full async surface live against OCI GenAI (us-chicago-1) on a prerelease `oci` SDK build, with `OCIAsyncClient._native` confirmed to be the SDK's `AsyncGenerativeAiInferenceClient` (not the fallback) in every case:

| Model | `ainvoke` | `astream` | tool call |
|---|---|---|---|
| meta.llama-3.3-70b-instruct | ✅ (+usage) | ✅ 12 chunks | ✅ |
| cohere.command-a-03-2025 | ✅ (+usage) | ✅ 7 chunks | ✅ |
| google.gemini-2.5-flash | ✅ (+usage) | ✅ 4 chunks | ✅ |
| openai.gpt-4o | ✅ (+usage) | ✅ 12 chunks | ✅ |
| openai.gpt-5 (non-default `temperature`) | ✅ — the 400 param-compat retry engages through the translated `ServiceError` and succeeds | ✅ 12 chunks | n/a |
| xai.grok-4.20-reasoning | ✅ | ✅ 28 chunks incl. **18 `reasoning_content` deltas** | n/a |

Also verified:

- **3 concurrent `astream` calls on one shared `ChatOCIGenAI` instance** — no cross-contamination (each stream returned its own token), exercising the shared native aiohttp session under concurrency.
- **Embeddings**: `cohere.embed-english-v3.0` (1024-dim) and `cohere.embed-v4.0` (1536-dim) via `aembed_documents` + `aembed_query`.
- **UTF-8 SSE decoding**: streamed multibyte content (`café ☕ 日本語`) arrives intact.
- **Sync/async parity on failures**: model ids the tenancy doesn't currently serve (`xai.grok-3`, `xai.grok-4`, on-demand `rerankText`) return the same 404 through the sync SDK client and through this async path — surfaced per contract as `OCIAsyncRequestError(404, <OCI error envelope>)`.

fede-kamel:
## Review guide — files changed

**[`libs/oci/langchain_oci/common/async_support.py`](https://github.com/oracle/langchain-oracle/pull/300/files#diff-087c518b27bfbc85728ace8226a001268e6799f5518929380b533d6ce3c9106c)** (+154/−5) — the entire behavior change lives here:
- [`_load_native_async_client_cls()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L37) — feature detection; `ImportError`/`AttributeError` on released SDKs → fallback.
- [`_service_error_body()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L51) — rebuilds the `{"code", "message"}` OCI envelope from `ServiceError` so `OCIAsyncRequestError.body` is transport-independent (this is what keeps the [`param_compat` 400 retry](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/common/param_compat.py#L116) working).
- [`OCIAsyncClient._build_native_client()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L135) — constructs the native client with the sync client's `config`/`signer`/`service_endpoint` and `skip_deserialization=True`; any construction failure logs at debug and falls back.
- [`_native_call()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L161) / [`_native_stream()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L178) — the two delegation primitives (non-streaming ops via generated methods; SSE via `AsyncBaseClient.call_api_stream`, since the generated ops are non-streaming only).
- [`chat_async`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L345) / [`embed_text_async`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L395) / [`rerank_text_async`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L429) / [`generate_text_async`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L463) — each gains a `if self._native is not None:` branch at the top; the fallback code below each branch is byte-for-byte unchanged.
- [`close()`](https://github.com/oracle/langchain-oracle/blob/94b540e48552c2ca2898bf9b9306086cea6c70c6/libs/oci/langchain_oci/common/async_support.py#L229) — also closes the native client.

**[`libs/oci/tests/unit_tests/common/test_native_async_client.py`](https://github.com/oracle/langchain-oracle/pull/300/files#diff-18a93ac3309b5a0e73b6e8b933e681f14dc3ffc290b3a54fbe0a4b9f0d77baa5)** (+378, new) — 15 tests: constructor wiring, delegation for every op (invoke + stream), `ServiceError` → `OCIAsyncRequestError` translation (incl. the double-encoded gpt-5-style 400 round-tripping through [`adjust_request_for_param_error`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/common/param_compat.py#L116)), fallback when the class is missing or construction raises, and `ChatOCIGenAI._agenerate`/`_astream` end-to-end over a fake native client.

**[`libs/oci/tests/unit_tests/common/__init__.py`](https://github.com/oracle/langchain-oracle/pull/300/files#diff-695b0f1893a408a206ac638db71de1dd04520f97f3317b7241a19f4ae6231df7)** (new, empty) — package marker for the new test directory.

**[`libs/oci/tests/unit_tests/chat_models/test_oci_responses_api.py`](https://github.com/oracle/langchain-oracle/pull/300/files#diff-cc18912d62283879d30d76ad397475b2eb50b7e42e671e4bcf7df8b37591ccf6)** (+3/−3) — drive-by lint fix, no behavior change: three `var-annotated` mypy errors that pre-dated this PR (came in with #242) and were failing `make lint` on main; annotated the dict literals so CI lint is green again.

Not touched by design (consume `OCIAsyncClient` unchanged): [`chat_models/async_mixin.py`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/chat_models/async_mixin.py), [`embeddings/oci_generative_ai.py`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/embeddings/oci_generative_ai.py), [`document_compressors/oci_generative_ai_rerank.py`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/document_compressors/oci_generative_ai_rerank.py), [`llms/oci_generative_ai.py`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/langchain_oci/llms/oci_generative_ai.py), [`pyproject.toml`](https://github.com/oracle/langchain-oracle/blob/main/libs/oci/pyproject.toml) (no dependency changes).

fileames:
Nice direction, 

Some things noticed, not all may be blocking:  

1. `_service_error_body` discards the structured error payload (common/async_support.py:59)

` return json.dumps({"code": error.code, "message": error.message})`
 For OpenAI-style bodies — `{"error": {"param": "temperature", "code": "unsupported_value"}} `— there is no top-level code or message, so `base_client.get_deserialized_service_code_and_message` returns None for both. The real payload survives only via deserialized_data, which ServiceError.__init__ merges into args[0]. This repo already documents exactly that, in `common/param_compat.py`

>  the OCI SDK does not promote the nested payload into ServiceError.message (which is then None); the deserialized body is merged into the exception's args[0] details dict instead

So on the native path a gpt-5 400 becomes `body == '{"code": null, "message": null}' `→ `extract_unsupported_param` returns (None, None) → `adjust_request_for_param_error` returns False → `_agenerate/_astream` re-raise instead of retrying. That's the behavior the PR description says is unchanged. Every error with a non-standard envelope also loses all of its diagnostic text.

Suggested fix — serialize the details dict, which is JSON-serializable and already carries code/message plus the merged body. Alternatively, pass the ServiceError itself through — extract_unsupported_param already accepts one.

2. The native client is constructed without timeout (common/async_support.py:147)

`_build_native_client` forwards only config / signer / service_endpoint / skip_deserialization. The generated client sets `base_client_init_kwargs['timeout']` only when 'timeout' in kwargs; otherwise BaseClient falls back to `(DEFAULT_CONNECTION_TIMEOUT, DEFAULT_READ_TIMEOUT) = (10.0, 60.0)`.

That's the exact default this repo overrides on purpose, per common/auth.py:

> Overrides the OCI SDK default of (10, 60) because long-context LLM generation and streaming calls routinely exceed 60s read timeout.

Concretely: `await llm.ainvoke(long_prompt)` against a long-context model that takes >60s used to succeed under the 300s aiohttp budget, and now fails with a read timeout — after which `DEFAULT_RETRY_STRATEGY` re-issues the expensive request. And a user who sets `OCI_REQUEST_TIMEOUT=600`, or `passes timeout=`, gets no effect on the async path at all.

Relatedly, the class docstring says "the native client uses the SDK's client-level timeout (same 300s default)" — that isn't accurate as written; nothing passes 300s (or the repo's 240s).

3. The param-compat test encodes a ServiceError shape the SDK never produces (tests/unit_tests/common/test_native_async_client.py:248)

`native.error = ServiceError(400, "400", {}, inner)
`

Rebuilding this test with `deserialized_data=` makes it fail against the current implementation.

4. operation_name doesn't match SDK convention (common/async_support.py:485)

The generated ops use snake_case: "generate_text", "embed_text", "rerank_text", "chat". "chat" at line 374 is already right; "generateText" should be "generate_text". It's diagnostics-only (it lands in ServiceError.operation_name and the troubleshooting link), but async failures will otherwise report a different operation name than sync for the same call.

Separately: `_native_stream` calls `call_api_stream` with no `retry_strategy`, while the generated non-streaming ops apply `retry.DEFAULT_RETRY_STRATEGY`. So the "retry-strategy parity with the sync client" in the description holds for invoke but not for streaming.

5. `close()` leaves `self._native` set (common/async_support.py:231)

The fallback transport self-heals — `_get_session` recreates a closed session — and the docstring advertises async with usage. After `__aexit__`, reusing the same `OCIAsyncClient` hits an already-closed native client instead of reconnecting, and a second `close() `closes it twice. The in-repo `aclose()` helpers drop the cached property, so nothing breaks in-tree today; setting `self._native = None` after closing restores parity.

fede-kamel:
Thanks @fileames — all five points addressed in the latest commit, plus the floor bump now that the client is public.

1. **Error envelope.** `_service_error_body` now serialises the `ServiceError` details dict (which is where the SDK merges `deserialized_data`) with `code`/`message` filled in as fallbacks, instead of only `code`/`message`. Nested OpenAI-style `{"error": {...}}` bodies reach `extract_unsupported_param` intact, and nothing the service returned is dropped. Live, a gpt-5 400 now surfaces as `{"target_service": ..., "status": 400, "code": "400", "opc-request-id": ..., "message": "Invalid 'maxTokens' ...", "operation_name": "chat", ...}` rather than two fields. The param-compat unit test was rebuilt with the real `deserialized_data=` shape as you suggested, and there is a new test asserting the full envelope.

2. **Timeout.** `OCIAsyncClient` takes `timeout` and resolves it with the same `_resolve_timeout` as the sync client (`OCI_REQUEST_TIMEOUT`, else `(10, 240)`), forwards it to the native client at construction, and every consumer passes its sync client's `base_client.timeout`. Verified live: default → native `(10, 240)`; `OCI_REQUEST_TIMEOUT=600` → sync `(10, 600)` and native `(10, 600)`. Fallback per-call timeouts default to the same read timeout. The "same 300s default" docstring is gone. One data point: on 2.185.x `AsyncBaseClient` defaults to 300, not `(10, 60)`, but the override gap you described was real and is closed.

3. **Test shape.** Done, see 1.

4. **Operation name / retry.** `generate_text` (snake case). `_native_stream` now resolves the retry strategy exactly like the generated ops (operation → client → `DEFAULT_RETRY_STRATEGY`) and passes it to `call_api_stream`, with tests for both the default and a client-level strategy.

5. **`close()`.** Drops `self._native`; the next call rebuilds it lazily, so `async with` reuse and a second `close()` both work. Covered by tests, including that a client that never had a native class does not get one revived.

Also: `oci>=2.185.0` is now the floor (the public release that ships `AsyncGenerativeAiInferenceClient`). I kept the aiohttp path only as a construction-failure safety net and would remove it in a follow-up once this settles, to keep this diff reviewable.

Verification on this commit: unit suite 527 passed / 47 skipped against oci 2.185.2, `make lint` clean; live in us-chicago-1 on the public SDK: native path confirmed for meta.llama-3.3-70b, cohere.command-a-03-2025, google.gemini-2.5-flash, openai.gpt-4o and xai.grok-4.20 (ainvoke, astream, tool call), 3 concurrent streams on one instance, cohere embed v4.0/v3.0, and the gpt-5 temperature retry.

fede-kamel:
One amendment to the floor bump above: `oci>=2.185.0` cannot be the floor for the whole Python range, because oci 2.185.x requires pyOpenSSL 26.2, whose cryptography dependency excludes Python 3.9.0 and 3.9.1, and the lock would not resolve. The pin is now split by interpreter: `oci>=2.185.0` on Python 3.10+, `oci>=2.173.0` on 3.9, with the lock regenerated accordingly (oci 2.185.2 resolves on 3.10+, 2.173.0 stays for 3.9). That also means the aiohttp fallback is still needed on 3.9 rather than being dead code, so it stays until 3.9 support is dropped. Native-async unit tests and lint are green on the amended head.

fileames (approved):

PR #301, opened 2026-08-19
feat(js/langchain-oracledb): add end-to-end RAG pipeline example

sudarshan12s:
This PR adds a complete end-to-end RAG example demonstrating how to build a retrieval-augmented generation pipeline using Oracle Database AI Vector Search, LangChain.js, and OCI Generative AI.

The example covers the full RAG flow:

`Load documents → Split into chunks → Generate embeddings → Store/search vectors → Build prompt → Generate answer with OCI Generative AI`


sharadraju (commented):
LGTM

sharadraju (approved):
LGTM

fede-kamel (changes requested):
Nice, readable end-to-end sample — the staged retrieval/prompt/generation output is a great teaching format, and I verified the defaults are real (`xai.grok-4.3` is in the current on-demand catalog and the endpoint resolves). One real bug though, plus a few smaller items.

**Bug: deterministic chunk IDs collide across files → silent data loss at ingestion.**

`OracleDocLoader` in `dir` mode builds `metadata` *only* from the `<meta>` tags returned by `DBMS_VECTOR_CHAIN.UTL_TO_TEXT` (see `_extract` in `libs/js/langchain-oracledb/src/document_loaders.ts`) — it never sets `metadata.source`, and it can't, because the DB function only receives the file's BLOB, never its path. So in this sample:

- `String(doc.metadata.source ?? "doc")` evaluates to `"doc"` for **every** document, and `sourceFile` becomes `"doc"` too (the retrieval table will show `doc (Chunk #1)` for everything).
- `makeStableChunkId("doc", 1)`, `("doc", 2)`, … produce **identical IDs across files**. With `mutateOnDuplicate: true`, each file upserts over the previous file's chunks.
- With the two shipped example files (each small enough to be a single 200-word chunk), the vector store ends up containing **only `technology.txt`** — `movies.txt` is silently overwritten. The demo question happens to target the surviving file, so the bug is invisible until someone asks about movies and gets "I don't have enough information…".

Suggested fix — simplest first:

1. **Drop the deterministic-ID + `mutateOnDuplicate` machinery entirely.** The table is `dropTablePurge`d immediately before ingestion in REBUILD mode, and REUSE mode never ingests, so stable IDs buy nothing here — let `OracleVS` generate IDs.
2. If you want to keep stable IDs for illustration, make them actually unique: load each file individually (`new OracleDocLoader(conn, { file })` over the directory listing) and attach the real filename to `metadata` yourself — that also fixes the `Source` column display — or at minimum hash a global chunk counter instead of the per-file one.

**Smaller items:**

- README env-var table is missing `RESET_VECTOR_STORE` and `OCI_MAX_TOKENS`, both of which the script reads.
- README: "- `testuser` in this case" in the `ORACLEDB_USER` row reads like a leftover from the blog draft.
- `@oracle/langchain-oracledb` is pinned to exactly `1.0.0` while the repo is about to release 1.6.0 (#292) — suggest `^1.0.0` so the sample picks up fixes (e.g. the sparse-vector fix in #298).
- Missing newline at end of `oracle-rag-full-pipeline.ts`.
- Non-blocking / follow-up: now that #285 is merged, the ~120 lines of hand-rolled OCI SDK client + `chatWithOci` plumbing could become a few lines of `OciGenAiGenericChat` from this repo's own `@oracle/langchain-oci`. It isn't published to npm yet, so it can't be used in an installable sample today — but worth a TODO so the sample dogfoods the LangChain integration once it's published.

fede-kamel:
Pushed commit 55a0cf4c to this branch (maintainer edit) applying the review findings, so you don't have to redo the analysis:

- **Chunk-ID collision fixed**: documents are now loaded per file with the real filename attached to `metadata.source` (dir-mode extraction happens inside the database, which never sees the path). The deterministic-ID/`mutateOnDuplicate` machinery is removed entirely — the table is purged right before ingestion and the REUSE path never ingests, so OracleVS-generated IDs are sufficient and nothing can silently overwrite another file's chunks. The retrieval table now shows real filenames too.
- **`npm run typecheck` now passes** (it failed on the original code): narrowed `client.chat()`'s `ChatResponse | ReadableStream` union for the non-streaming call, and added `@types/oracledb` (node-oracledb ships no type declarations, so `strict` mode failed on the import).
- README: documented `RESET_VECTOR_STORE` and `OCI_MAX_TOKENS`, removed the `testuser` leftover.
- `@oracle/langchain-oracledb` pinned as `^1.0.0`.

Verified with `npm install && npm run typecheck` against the published packages. Please double-check you're happy with the per-file loading approach — feel free to adjust.

fede-kamel (dismissed):
Approving with the fixes from 55a0cf4c applied — see the comment for details. If you'd rather solve the ID collision differently, happy to re-review.

sudarshan12s:
> * npm run typecheck

Thanks @fede-kamel so much for pointing out the issues. Yes your patch looks perfect.

sudarshan12s:
> Nice, readable end-to-end sample — the staged retrieval/prompt/generation output is a great teaching format, and I verified the defaults are real (`xai.grok-4.3` is in the current on-demand catalog and the endpoint resolves). One real bug though, plus a few smaller items.
> 
> **Bug: deterministic chunk IDs collide across files → silent data loss at ingestion.**
> 
> `OracleDocLoader` in `dir` mode builds `metadata` _only_ from the `<meta>` tags returned by `DBMS_VECTOR_CHAIN.UTL_TO_TEXT` (see `_extract` in `libs/js/langchain-oracledb/src/document_loaders.ts`) — it never sets `metadata.source`, and it can't, because the DB function only receives the file's BLOB, never its path. So in this sample:
> 
> * `String(doc.metadata.source ?? "doc")` evaluates to `"doc"` for **every** document, and `sourceFile` becomes `"doc"` too (the retrieval table will show `doc (Chunk #1)` for everything).
> * `makeStableChunkId("doc", 1)`, `("doc", 2)`, … produce **identical IDs across files**. With `mutateOnDuplicate: true`, each file upserts over the previous file's chunks.
> * With the two shipped example files (each small enough to be a single 200-word chunk), the vector store ends up containing **only `technology.txt`** — `movies.txt` is silently overwritten. The demo question happens to target the surviving file, so the bug is invisible until someone asks about movies and gets "I don't have enough information…".
> 
> Suggested fix — simplest first:
> 
> 1. **Drop the deterministic-ID + `mutateOnDuplicate` machinery entirely.** The table is `dropTablePurge`d immediately before ingestion in REBUILD mode, and REUSE mode never ingests, so stable IDs buy nothing here — let `OracleVS` generate IDs.
> 2. If you want to keep stable IDs for illustration, make them actually unique: load each file individually (`new OracleDocLoader(conn, { file })` over the directory listing) and attach the real filename to `metadata` yourself — that also fixes the `Source` column display — or at minimum hash a global chunk counter instead of the per-file one.
> 
> **Smaller items:**
> 
> * README env-var table is missing `RESET_VECTOR_STORE` and `OCI_MAX_TOKENS`, both of which the script reads.
> * README: "- `testuser` in this case" in the `ORACLEDB_USER` row reads like a leftover from the blog draft.
> * `@oracle/langchain-oracledb` is pinned to exactly `1.0.0` while the repo is about to release 1.6.0 (#292) — suggest `^1.0.0` so the sample picks up fixes (e.g. the sparse-vector fix in #298).
> * Missing newline at end of `oracle-rag-full-pipeline.ts`.
> * Non-blocking / follow-up: now that #285 is merged, the ~120 lines of hand-rolled OCI SDK client + `chatWithOci` plumbing could become a few lines of `OciGenAiGenericChat` from this repo's own `@oracle/langchain-oci`. It isn't published to npm yet, so it can't be used in an installable sample today — but worth a TODO so the sample dogfoods the LangChain integration once it's published.

We are planning to upload npm module for oci, i will make a note to update this example. 

Thanks,

fede-kamel:
@sudarshan12s the September 7 switch to `OciGenAiGenericChat` was the right move, but as pushed the sample did not build: `@oracle/langchain-oci` was never added to the sample's `package.json` (fresh `pnpm install` then `pnpm typecheck` fails with TS2307), and the `@langchain/core` pin at 1.1.26 lacks the `language_models/structured_output` export the package needs, so it also failed at runtime with `ERR_PACKAGE_PATH_NOT_EXPORTED`. I pushed f0771f37 onto your branch: adds `@oracle/langchain-oci@^0.0.1`, bumps `@langchain/core` to `^1.2.10`, and drops the two raw OCI SDK packages that are no longer imported.

Verified end to end against Oracle AI Database 26ai Free and OCI GenAI: ingestion of the two example documents into OracleVS with the IVF index, retrieval, prompt construction and the final cited answer from `xai.grok-4.20` all complete, typecheck clean. One note for later: the sample hardcodes `OciGenAiNewClientAuthType.ConfigFile`, so it cannot run with a session-token profile; the package's `Session` auth type works, so an `OCI_AUTH_TYPE` switch would make the sample usable for everyone.

The chunk-ID fix from August is in. Re-approving and merging once CI is green.

fede-kamel (approved):
Re-approved after pushing f0771f37 (declare @oracle/langchain-oci, require @langchain/core >=1.2, drop unused raw SDK deps). Verified end to end against Oracle AI Database 26ai and OCI GenAI: ingestion into OracleVS with IVF index, retrieval, prompt construction and the final cited answer from xai.grok-4.20; typecheck clean. The August chunk-ID fix is included.

PR #304, opened 2026-09-09
fix(oci): skip the `[DONE]` SSE sentinel in sync streaming (ChatOCIGenAI, OCIGenAI)

fede-kamel:
## Problem

OCI GenAI now closes chat streams for **Meta Llama** models with an OpenAI-style terminal frame:

```
data: [DONE]
```

`ChatOCIGenAI._stream` (and `OCIGenAI._stream` for the completion LLM) call `json.loads(event.data)` on every SSE frame, so every sync `.stream()` call on those models raises after the last real chunk:

```
json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)   # e.doc == '[DONE]'
```

Reproduced on `main` with nothing more than `ChatOCIGenAI(model_id="meta.llama-3.3-70b-instruct", ...).stream("Say hello")`. Same for `meta.llama-4-maverick-17b-128e-instruct-fp8`. Cohere, Gemini, OpenAI and xAI streams do not carry the sentinel and are unaffected. `.astream()` already worked because the async transport skips `[DONE]` (`common/async_support.py`), as does `_stream_responses_api`.

## Fix

- `ChatOCIGenAI._stream`: skip frames whose payload is empty or `[DONE]` before `json.loads` (small `_is_sse_sentinel` helper).
- `OCIGenAI._stream` (completion LLM): same guard.
- Behaviour is otherwise unchanged; no new chunks are emitted for the skipped frames.

## Tests

- `tests/unit_tests/chat_models/test_oci_generative_ai.py::TestStreamDoneSentinel` — stream with a trailing `[DONE]`, and with empty / padded / `None` frames, yields the same chunks as a clean stream.
- `tests/unit_tests/llms/test_stream_done_sentinel.py` — same for `OCIGenAI.stream()`.
- Full `libs/oci` unit suite: 505 passed, 47 skipped. `ruff check`, `ruff format --diff`, `mypy` clean.

## Live verification (OCI GenAI, us-chicago-1)

| Model | `.stream()` on main | with this PR |
|---|---|---|
| meta.llama-3.3-70b-instruct | JSONDecodeError on `[DONE]` | OK |
| meta.llama-4-maverick-17b-128e-instruct-fp8 | JSONDecodeError on `[DONE]` | OK |
| cohere.command-a-03-2025, google.gemini-2.5-flash, openai.gpt-4o, openai.gpt-5, xai.grok-4.20 | OK | OK |

This is also what was making the langchain-tests conformance suite in #277 red: with this guard applied, that suite goes from 10 failed / 27 passed to 37 passed / 11 skipped / 2 xfailed / 0 failed against the live service.

### DIFF
--- a/libs/oci/langchain_oci/llms/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/llms/oci_generative_ai.py
@@ -357,6 +357,10 @@ class OCIGenAI(LLM, OCIGenAIBase):
         response = self.client.generate_text(invocation_obj)
 
         for event in response.data.events():
+            # Skip the non-JSON ``data: [DONE]`` terminal frame / empty frames
+            # (see ChatOCIGenAI._stream) instead of raising JSONDecodeError.
+            if event.data is None or event.data.strip() in ("", "[DONE]"):
+                continue
             json_load = json.loads(event.data)
             if "text" in json_load:
                 event_data_text = json_load["text"]
--- a/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
+++ b/libs/oci/langchain_oci/chat_models/oci_generative_ai.py
@@ -110,6 +110,18 @@ def _build_chat_completions_headers(compartment_id: str) -> Dict[str, str]:
     return {COMPARTMENT_ID_HEADER: compartment_id}
 
 
+def _is_sse_sentinel(data: Optional[str]) -> bool:
+    """Return True for SSE frames that carry no JSON payload.
+
+    The OCI GenAI streaming endpoint emits a terminal ``data: [DONE]`` frame
+    for some models (seen live on ``meta.llama-3.3-70b-instruct`` and
+    ``meta.llama-4-maverick-17b-128e-instruct-fp8`` in September 2026).
+    Empty frames are treated the same way so keep-alives never reach
+    ``json.loads``.
+    """
+    return data is None or data.strip() in ("", "[DONE]")
+
+
 class ChatOCIGenAI(ChatOCIGenAIAsyncMixin, BaseChatModel, OCIGenAIBase):
     """ChatOCIGenAI chat model integration.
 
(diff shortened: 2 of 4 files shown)


fede-kamel:
@fileames could you do one review pass over the open set when you have a slot? Everything below is green on OCA and CI and has been exercised end to end against OCI GenAI and Oracle AI Database 26ai this week.

- **#304** (this one), small but urgent: the service now ends Llama streams with `data: [DONE]` and `ChatOCIGenAI.stream()` crashes on `main`. One guard plus tests. Please take this first.
- **#298**: sudarshan's two nits are applied; sharadraju is waiting on the approval to include it in the next JS release.
- **#293**: replied to your comments, pushed the quoting and NaN fixes; the narrowing-semantics follow-up is coming as a separate PR with a release note, as you suggested.
- **#277**: rebased. Its 10 red conformance tests are the #304 bug; with #304 applied it is 37 passed / 0 failed live.
- **#300**: all five of your points addressed, replied point by point; native path verified on the public SDK, fallback path verified on Python 3.9.
- **#292** and **#289**: no review yet. #292 is the 1.6.0 bump for three unreleased features; #289 is the deepagents docs set with the sample verified live.

Thanks!

libs/oci/langchain_oci/llms/oci_generative_ai.py
```diff
@@ -357,6 +357,10 @@ def _stream(
         response = self.client.generate_text(invocation_obj)
 
         for event in response.data.events():
+            # Skip the non-JSON ``data: [DONE]`` terminal frame / empty frames
+            # (see ChatOCIGenAI._stream) instead of raising JSONDecodeError.
+            if event.data is None or event.data.strip() in ("", "[DONE]"):
```
fileames, line 362: Should we duplicate this here?

fileames (approved):
Just left a minor comment, otherwise looks good

libs/oci/langchain_oci/llms/oci_generative_ai.py
fede-kamel: Fair point, it is the same guard twice. I kept the legacy `OCIGenAI` path free of an import from `chat_models` to avoid a cross-module dependency, but the cleaner shape is to move `_is_sse_sentinel` into `common/utils.py` and import it from both stream loops. Since pushing here would dismiss the approval, I will do that as a small follow-up PR right after this merges, unless you would rather it land in this one.

fileames: I can reapprove after the change

fede-kamel: Follow-up is up: #313 moves the guard to `langchain_oci.common.utils.is_sse_sentinel` and imports it from both loops.

Issue #265, opened 2026-07-22
feat(oci): expose logprobs in ChatOCIGenAI


fede-kamel:
## Motivation

Logprobs are used for confidence scoring, hallucination detection, and eval pipelines. `ChatOCIModelDeploymentVLLM`/`TGI` already expose `logprobs`/`top_logprobs` params and parse them into `generation_info`, but the flagship `ChatOCIGenAI` has no logprobs support at all, even for provider paths where the underlying API can return them (OpenAI-family and Meta models on OCI GenAI).

Peer precedent: `ChatOpenAI` exposes `logprobs`/`top_logprobs` and returns them in `response_metadata`.

## Current state

No `logprobs` handling anywhere in `libs/oci/langchain_oci/chat_models/oci_generative_ai.py` or its provider modules.

## Proposed scope

- Accept `logprobs` / `top_logprobs` model kwargs on `ChatOCIGenAI` for provider paths that support them; surface results in `response_metadata` (and per-chunk `generation_info` when streaming).
- Raise a clear error (or document no-op) for providers that don't support logprobs (e.g. Cohere).

## Acceptance criteria

- `ChatOCIGenAI(model_kwargs={"logprobs": True})` returns logprobs in `response_metadata` for supported models, in both invoke and stream.
- Unit tests per provider path.

PR #306, opened 2026-09-10
fix(oci): pin openai below 3

fede-kamel:
## Problem

`libs/oci` allows `openai>=2.6.1` with no upper bound. openai 3.x changed how a custom httpx `auth` object is accepted, and rejects oci-openai's `OciSessionAuth` with `TypeError: Invalid "auth" argument`. A fresh `pip install langchain-oci` today resolves openai 3.x, so the Responses API path (`ChatOCIGenAI(use_responses_api=True)`, `ChatOCIOpenAI`) is broken for new installs while CI, which runs the lock, stays green.

## Fix

Pin `openai>=2.6.1,<3` and regenerate the lock. Nothing else changes.

## Verification

- openai 3.13.0: 21 of the 26 Responses API unit tests fail with the TypeError above.
- With the pin (openai 2.x): full `libs/oci` unit suite 502 passed, 47 skipped.
- `poetry check --lock` clean.

The pin can be lifted once oci-openai supports the openai 3.x auth contract.

### DIFF
--- a/libs/oci/poetry.lock
+++ b/libs/oci/poetry.lock
@@ -5601,4 +5601,4 @@ deepagents = ["deepagents", "langchain-oracledb", "langgraph-oracledb", "opensea
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.9,<4.0"
-content-hash = "9aebd96c8d65d23baf65a40c26d2bc17cf75bf1d54b412ba04afc32fb5a557b6"
+content-hash = "f0e93ee35d6b83cb620926fc9ebc0c645f577084b0611c86e051580cfebad2d1"
--- a/libs/oci/pyproject.toml
+++ b/libs/oci/pyproject.toml
@@ -13,7 +13,9 @@ dependencies = [
     "oci>=2.173.0",
     "pydantic>=2,<3",
     "aiohttp>=3.12.14",
-    "openai>=2.6.1",
+    # openai 3.x changes how a custom httpx auth is accepted; oci-openai's
+    # OciSessionAuth is rejected with TypeError: Invalid "auth" argument.
+    "openai>=2.6.1,<3",
     "oci-openai>=1.0.0",
     "langchain-openai>=0.3.35,<1.0.0; python_version < '3.10'",
     "langchain-openai>=1.1.0,<2.0.0; python_version >= '3.10'",


fede-kamel:
@fileames thanks for the quick turnaround on #304 and #313. With those in, the remaining open PRs are all green (CI + OCA), conflict-free against `main`, and verified end to end; each just needs your approval. Suggested order, smallest first:

1. #306 `fix(oci): pin openai below 3` (2 files) — the openai 3.x line breaks the Responses API tests.
2. #311 `build(langgraph-oracledb): bump pytest-asyncio to 1.x` (2 files) — pytest 9 crashes collection on 0.23; this is what blocked the langgraph-oracledb 1.0.2 release run, so I will re-run the release right after it lands.
3. #292 `chore: bump langchain-oracledb to 1.6.0` (1 file).
4. #309 `docs(js): session-auth switch for the RAG sample` (3 files) — Sudarshan already gave it a thumbs up.
5. #307 `test(oci): refresh deepagents backend and xAI integration tests` (4 files).
6. #310 `feat(oci): expose token log probabilities in ChatOCIGenAI` (5 files) — closes #265.
7. #300 `feat(oci): native AsyncGenerativeAiInferenceClient` (9 files) — all five points from your August review are addressed; `oci>=2.185.0` on Python 3.10+, fallback kept for 3.9.
8. #293 `feat(langgraph-oracledb): containment matching` (4 files) — #308 (strict filter leaves) is stacked on it and will come out of draft once it merges.
9. #289 `docs(oci): deepagents documentation set` (10 files, docs only).
10. #277 `test(oci): langchain-tests conformance suite` (4 files).

Happy to split or reorder if any of them needs a closer look.

fede-kamel:
@fileames following up on the list above. Everything is still green (CI + OCA) and conflict-free against `main`, and nothing has changed on any branch since 09-10, so whatever you already looked at still applies. To keep this from being a 10-PR wall, here are three ways to slice it. Pick whichever fits your time and I will sequence the rest around it.

**Option A: quick wins (three PRs, one to three files each)**
- #306 pin openai below 3
- #311 pytest-asyncio 1.x, which unblocks the langgraph-oracledb 1.0.2 release run
- #309 JS RAG sample session auth (Sudarshan and Sharad already gave it a thumbs up)

**Option B: A plus the two re-reviews where your comments were addressed**
- #300 native async client: all five points from your August review are in, with `oci>=2.185.0` on 3.10+ and the aiohttp fallback kept on 3.9
- #293 containment metadata filters: dotted keys and NaN handling fixed per your notes, verified live on 26ai. #308 comes out of draft once this lands.

**Option C: split by depth**
- You take a close look at #300, #293, #310, and #277, since those touch runtime code.
- #292, #307, #289, and #306 are a version bump, tests, docs, and a dependency pin, all verified end to end, so a lighter pass is enough there.

If a 30-minute walkthrough would be faster than reading, I am happy to set one up. And if any single PR needs splitting further, say which and I will do it.

fileames (approved):

PR #307, opened 2026-09-10
test(oci): refresh deepagents backend and xAI integration tests for the current APIs

fede-kamel:
Two groups of integration tests on `main` fail for reasons unrelated to the code under test, which makes every live run look worse than it is.

**deepagents 0.7 removed backend factories.** `TestDeepAgentBackend` (3 tests) and `TestDeepAgentAllParamsCombined` still pass the `StateBackend` class or `lambda rt: StoreBackend(rt)` factories and fail with `TypeError: backend must be an initialized backend instance. Backend factories were removed in deepagents 0.7`. They now pass `StateBackend()`, `StoreBackend(namespace=...)`, and an explicit `CompositeBackend` in place of the factory-lambda case. The `deepagents` extra floor moves from 0.6.1 to 0.7.0 to match what the tests and the docs (#289) describe; lock regenerated.

**`xai.grok-3-mini-fast` is no longer served on-demand** (404 from the service, model still listed in the catalog), so `test_multi_model_tool_calling` and `test_multi_model_support` carried a guaranteed failure. Replaced with `xai.grok-4.20`, which is served.

Verification: all six affected tests pass live against OCI GenAI (us-chicago-1) with deepagents 0.7.13; `ruff check` and `ruff format` clean.

### DIFF
--- a/libs/oci/poetry.lock
+++ b/libs/oci/poetry.lock
@@ -185,33 +185,34 @@ files = [
 
 [[package]]
 name = "anthropic"
-version = "0.97.0"
+version = "1.5.0"
 description = "The official Python library for the anthropic API"
 optional = true
-python-versions = ">=3.9"
+python-versions = ">=3.10"
 groups = ["main"]
 markers = "python_version >= \"3.11\" and python_version < \"3.14\" and extra == \"deepagents\""
 files = [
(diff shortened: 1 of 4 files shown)


fede-kamel:
@fileames test-only PR: refreshes the deepagents backend and xAI integration tests for the current APIs. Rebased on main yesterday after #306 landed (the only conflict was poetry.lock), CI 13/13. A quick pass would be great.

fileames (approved):

PR #303, opened 2026-08-24
fix(js/oci): Address issues related to regionId and structured output flow.

sudarshan12s:
### Fixes:

- Remove hardocde regionId (Chicago) and leave it undefined if its not provided which causes default region taken fro m ~/.oci/config
- Added structured output via forced function calling and LangChain’s schema-validating parser.
- Supports includeRaw; clearly rejects unsupported `strict` , `jsonMode` and `jsonSchema` adapter modes.
- Improved `bindTools`() validation by ensuring that a named `tool_choice` refers to one of the tools being bound, preventing invalid tools and `toolChoice` combinations from being sent to OCI. Added the test `OCI GenAI Generic bindTools rejects a named choice for an unbound tool.`
- Clarify in README.md for SEARCH_DOCUMENT and SEARCH_QUERY. Fix links in README.md

### Other minor changes:

- updated README.md to proper versions.
- package.json to update oci sdk.
- fix lockfile

PR #309, opened 2026-09-10
docs(js): session-auth switch for the RAG sample; structured-output naming note

fede-kamel:
Two small follow-ups from the #301 and #303 reviews.

**RAG sample: `OCI_AUTH_TYPE`.** The sample hardcoded `OciGenAiNewClientAuthType.ConfigFile`, so it could only run with API-key credentials. A new `OCI_AUTH_TYPE` env var (`ConfigFile`, the default, or `Session`) selects the auth type, which lets the sample run with a profile created by `oci session authenticate`. Unknown values fail fast with a clear error. README table updated.

**libs/js/oci README: structured output naming.** `withStructuredOutput` binds a function named after the schema, or `extract` when the schema has no name. When the schema has a top-level field literally called `name`, Llama 3.3 was observed filling that field with the function name; GPT-4o was not. The README now recommends passing an explicit `name` in that case.

Verification: the sample type-checks and runs end to end with `OCI_AUTH_TYPE=Session` against Oracle AI Database 26ai Free and OCI GenAI (ingestion, IVF index, retrieval, cited answer from `xai.grok-4.20`); an invalid `OCI_AUTH_TYPE` is rejected with the intended message; `pnpm format:check` passes for the README.


sudarshan12s (approved):
Thanks for the change.

LGTM.

sharadraju (approved):

fileames (approved):

PR #312, opened 2026-09-14
Fix ESM and CommonJS declaration resolution for @oracle/langchain-oci

sudarshan12s:
### Summary

Align the ESM and CommonJS TypeScript declaration entrypoints published by `@oracle/langchain-oci`.

The package publishes both ESM and CommonJS runtime artifacts (`.js` / `.cjs`) and declaration artifacts (`.d.ts` / `.d.cts`). Previously, both module modes could resolve through the ESM declaration path, and generated CommonJS declarations retained relative `.js` module specifiers.

For consumers using NodeNext-style module resolution, this could cause the same `@langchain/core` classes to be resolved through different declaration identities. Types containing private or protected members could then be reported as unrelated even when the consumer has a single version of `@langchain/core` installed.

For example, downstream consumers could see errors such as:

```text
Type 'OciGenAiEmbeddings' is not assignable to type 'Embeddings<number[]>'.
Types of property 'caller' are incompatible.
Two different types with this name exist, but they are unrelated.
```

### Changes

* Route ESM consumers to `index.d.ts` / `index.js`.
* Route CommonJS consumers to `index.d.cts` / `index.cjs`.
* Generate `.d.ts` entrypoints from the ESM declaration tree and `.d.cts` entrypoints from the CommonJS declaration tree.
* Rewrite relative `.js` module specifiers in generated CommonJS declarations to `.cjs`, keeping the declaration graph aligned with the CommonJS runtime graph.
* Add ESM and CommonJS consumer type checks covering the public LangChain types exposed by the package.

The resulting package resolution is:

```text
ESM consumer
  ├── types   → index.d.ts  → dist/*.d.ts
  └── runtime → index.js

CommonJS consumer
  ├── types   → index.d.cts → dist/*.d.cts
  └── runtime → index.cjs
```

### Testing

Added consumer type tests for both ESM and CommonJS resolution. The tests verify that OCI implementations remain assignable to their corresponding `@langchain/core` types, including `OciGenAiEmbeddings`.

`skipLibCheck` is intentional for these focused consumer tests because the repository's current TypeScript version reports unrelated diagnostics from transitive type packages. The consumer assignments themselves remain type-checked: before the CommonJS declaration rewrite, the test reproduces the `OciGenAiEmbeddings` / `AsyncCaller` incompatibility; with this change, both ESM and CommonJS consumer checks pass.


fede-kamel:
Verified locally on head 54ace0e7 (pnpm 10.32.1, Node 24):

- `pnpm build` passes end to end, including the new `test:consumer-types` step.
- Generated `dist/*.d.cts` files now use `.cjs` relative specifiers (`from "./chat_models.cjs"` etc.); no `.js`-suffixed relative import remains in any `.d.cts`.
- Root entrypoints resolve as intended: `index.d.ts → ./dist/index.d.ts`, `index.d.cts → ./dist/index.d.cts`.
- Negative check: restoring the old flat `exports` shape (`types` / `import` / `require`) and re-running `test:consumer-types` reproduces exactly the reported failure on the CommonJS consumer (TS1479 plus the two "not assignable ... Two different types with this name exist" errors for `OciGenAiEmbeddings` and `OciGenAiGenericChat`); the new conditional `exports` makes it pass.
- 99 unit tests pass; lint and format:check clean.

Looks good to me.

fede-kamel (approved):
Verified locally (build, .d.cts specifier rewrite, consumer-type check with and without the fix, 99 unit tests, lint). Approving.

fede-kamel:
Merged, thanks. This bumps `@oracle/langchain-oci` to 0.0.2, so it needs a publish; npm lists @sharadraju as the package's only maintainer, so it has to come from that account (or add me as a maintainer and I will do it).

Pre-publish check from `main` at 011b481a, `npm pack --dry-run --ignore-scripts` after `pnpm build`:

- `oracle-langchain-oci-0.0.2.tgz`, 39 files, 190 kB unpacked
- root: `index.js`, `index.cjs`, `index.d.ts`, `index.d.cts`, `package.json`, `LICENSE`, `README.md`
- `dist/`: 8 each of `.js`, `.cjs`, `.d.ts`, `.d.cts`; the `.d.cts` files carry `.cjs` relative specifiers
- no `src/`, `tests/` or `scripts/` files in the tarball

One note for whoever publishes: the `prepack` hook runs `pnpm build`, which requires pnpm 10 or newer (pnpm 9 fails on the package-level `pnpm-workspace.yaml` with "packages field missing"). `npx -y pnpm@10.32.1` works if a newer pnpm is not installed globally.

sharadraju:
Thanks @fede-kamel I will upload the latest 0.0.2 version in npm.
Let me see if I can give you upload access in npm as well

fede-kamel:
Thanks @sharadraju, much appreciated. Upload access would be great for future bumps; I will send you my npm username separately.

